PR #188, opened 2015-01-07
added -m option to missing2ref

wkretzsch:
Added option to the plugin `missing2ref` that sets missing to major allele (also works for multiallelic sites).  Works together with -p option.

I also added a test VCF. The following command should give an almost identical (except for one comment line) VCF to `test/plugin-missing2ref.out.vcf`

````
bcftools plugin missing2ref test/plugin-missing2ref.vcf -- -p -m
````

I have not integrated this test into the testing framework, but I hope that would be easy for someone familiar with the code to achieve.

I tested this code against release 1.1.  Not sure if it'll work on the current `develop` branch

### DIFF
--- a/plugins/missing2ref.c
+++ b/plugins/missing2ref.c
@@ -25,17 +25,20 @@ DEALINGS IN THE SOFTWARE.  */
 #include <stdio.h>
 #include <stdlib.h>
 #include <htslib/vcf.h>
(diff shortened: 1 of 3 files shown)


pd3:
Thank you

wkretzsch:
np.

PR #194, opened 2015-01-27
index: Support for both CSIv1 and CSIv2

pd3:
See corresponding branch in htslib


mcshane:
Closing this as it is outdated and the CSI2 branch never went ahead in htslib.

PR #197, opened 2015-01-28
bcftools convert --hapsamp

wkretzsch:
Implementation of `bcftools convert --hapsamp`: conversion from VCF to WTCCC style haps/sample format (according to this spec: https://mathgen.stats.ox.ac.uk/genetics_software/shapeit/shapeit.html#hapsample). 

Tested the new option on `test/convert.vcf` and get expected results.

`--haploid2diploid` option runs, but have not tested the results to make sure they are correct.


wkretzsch:
the `vcf_to_hapsample` function was mostly copied from `vcf_to_haplegendsample`.  One might consider merging the two functions for ease of maintenance, as they both do very similar things...

wkretzsch:
Just saw that there are some testing scripts.  Will add them. Just a sec.

wkretzsch:
ok, now the `--hapsample` option is being tested in `test/test.pl`

pd3:
Thank you Warren, that's great. I'll merge and check sometime later if the code could be reused as you suggested. Cheers

PR #198, opened 2015-01-29
implemented --vcf-ids for convert --hapsample

wkretzsch:
Manually checked output of `bcftools convert --hapsample test/check.vcf --vcf-ids` and it looks like the IDs are being printed ok.


wkretzsch:
This commit is included in pull request https://github.com/samtools/bcftools/pull/198

PR #199, opened 2015-01-29
Added flags for bcftools convert --gensample and --hapsample

wkretzsch:
Implemented support for `--vcf-ids` flag using `--hapsample` and `--gensample` in `bcftools convert`.
Implemented support for `--chrom` flag using `bcftools convert --gensample`.  

This allows the creation of .gen files according to the `shapeit` specification at https://mathgen.stats.ox.ac.uk/genetics_software/shapeit/shapeit.html#gensample
and haps/sample files according to the `shapeit` specification at https://mathgen.stats.ox.ac.uk/genetics_software/shapeit/shapeit.html#hapsample

Created test cases for `--gensample` with `--vcf-ids` and `--chrom` flags.

Also added an email address for myself in the AUTHORS file


wkretzsch:
Ok, the `shapeit` spec requires chromosome identifiers to be integers, so this may break when chromosome identifiers are not integers.  However, I think this gets us most of the way there without too much hassle.

pd3:
Thank you

Issue #204, opened 2015-02-05
[1.2] make install does not install plugins


wookietreiber:
The `make install` target does not install plugins. On a fresh install with 1.2 I get:

```
$ bcftools plugin -lv

No functional bcftools plugins were found. The environment variable BCFTOOLS_PLUGINS is not set.
```

Is there a default installation location for these plugins, where these are looked up?

I suggest installing them to `$DESTDIR/$prefix/share/bcftools/plugins/*.so` and automatically including this path as the default search path for plugins (PR pending).

PR #206, opened 2015-02-05
make install installs plugins

wookietreiber:
`make install` now installs plugins to `$(DESTDIR)$(exec_prefix)/libexec/bcftools`, see #204.

What is still left as of now is that `bcftools plugins` searches this path for plugins automatically. This can be done, I guess, by appending this path to `BCFTOOLS_PLUGINS` when this environment variable is evaluated.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,7 @@ BGZIP  = $(HTSDIR)/bgzip
 TABIX  = $(HTSDIR)/tabix
 
 CC       = gcc
-CFLAGS   = -g -Wall -Wc++-compat -O2
+CFLAGS   = -g -Wall -Wc++-compat -O2 -DPKG_LIBEXEC_DIR="\"$(pkglibexecdir)\""
 DFLAGS   =
 OBJS     = main.o vcfindex.o tabix.o \
            vcfstats.o vcfisec.o vcfmerge.o vcfquery.o vcffilter.o filter.o vcfsom.o \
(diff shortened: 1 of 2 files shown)


jmarshall:
`pkgdatadir` is wrong for this because these host binaries are not architecture-independent data.

wookietreiber:
Thanks, @jmarshall, you are totally right.

pd3:
libexecdir directory does not exist on most linux distributions, `/usr/lib/package` is used instead in practice

Makefile
```diff
@@ -55,11 +55,12 @@ ifdef USE_GPL
...
 
-prefix      = /usr/local
-exec_prefix = $(prefix)
-bindir      = $(exec_prefix)/bin
-mandir      = $(prefix)/share/man
-man1dir     = $(mandir)/man1
+prefix        = /usr/local
+exec_prefix   = $(prefix)
+bindir        = $(exec_prefix)/bin
+pkglibexecdir = $(exec_prefix)/libexec/bcftools
```
jmarshall: You need to write a line in the Makefile,

    libexecdir = $(exec_prefix)/libexec

so that it can be overridden by the user with `make libexecdir=/somewhere/else`.  If you want, you can then define another variable in terms of it...

jmarshall:
As Petr notes, there is some disagreement among FHS folks about whether `libexec` is needed, and Debian/Ubuntu do not use it (but see [upcoming FHS changes in this matter](http://www.linuxbase.org/betaspecs/fhs/fhs/ch04s07.html)).  They would have the option of building with `make libexecdir=$(libdir)` as they do for other packages following the GNU conventions.

Or I would be happy enough with `$(libdir)/bcftools/plugins`.

Please create the pull request anew or squash further commits into the first one; that would be better for merging than one with false starts in it.

wookietreiber:
@pd3 I know. However, these distributions usually override this by explicitly specifying `libexecdir` or `pkglibexecdir` during `./configure` or `make` in their build instructions.

wookietreiber:
@jmarshall I'll squash at the end. I still would like to add this default path-to-be to vcfplugin.c.

wookietreiber:
Squashed. Now also appends pkglibexecdir to plugin_paths.

There may be one downside, though: you have to compile with all your paths set you want to use during install, e.g. the following does not work, because compile prefix which is used for pkglibexecdir is different than install prefix:

```
make -j4
make -j4 prefix=/tmp/foo install
```

Correct invocation would be:

```
make -j4 prefix=/tmp/foo
make -j4 prefix=/tmp/foo install
```

I know of no other way to get the path from the Makefile to the c source code.

wookietreiber:
I don't know if this downside works for you. I guess, you would have to update the installation instructions.

I don't know whether you want to switch to GNU autotools or cmake or some other build system in the near future? I think, these would also take care of this downside, because of the explicit advance configure step.

wookietreiber:
If you don't have any objections, I consider this ready to merge.

PR #207, opened 2015-02-08
view -x must be combined with -s or -S

tommycarstensen:
I was expecting to be able to use view --private without --samples, because my modified VCF only contains a single sample. To my surprise that was not the behaviour. Hence I suggest this edit to the docs for people like myself that make invalid assumptions.


mcshane:
Thanks Tommy.

Closed by edd56f75dcaa0d272ebc0734003eae2dfd72393a

PR #210, opened 2015-02-10
Printing accepted types when invalid type is given

tommycarstensen:
I can never remember, whether snps is in plural or not. And I can never remember, whether other is named "other" or "complex" or something third. With this pull request the error message is slightly more informative.


mcshane:
Thanks Tommy.

PR #217, opened 2015-02-20
-g miss and -g ^miss options not in error message

tommycarstensen:
The -g miss and -g ^miss options were not mentioned in the error message. As an aside it would be great, if there was also a nonrefhom option.


mcshane:
Thanks Tommy.

PR #223, opened 2015-02-25
Add LDFLAGS and LIBS variables to Makefile.

ghuls:
Add LDFLAGS and LIBS variables to Makefile so special linker flags and
custom library locations can be specified on "make" invocation.

Closes samtools/bcftools/#222


wookietreiber:
:+1:

jmarshall:
Note that this is incomplete (there are other link commands in the build system) and `$(LIBS)` needs to go at the end of the link command.  Thanks for the bug report; fixed by 54d6c828db53a3e13127599098fdb34985bf44ba.

PR #244, opened 2015-04-22
gvcf fill ref allele from fasta

### LINKED ISSUE #243: gvcf2vcf not filling in REF allele correctly
When fixing #242 noticed that `gvcf2vcf` is just copying the REF allele from the beginning of the block. Command needs to take a reference fasta as input to fill in REF correctly.

mcshane:
* when filling in gvcf reference blocks need the
  samtools faidx indexed reference fasta to fill in
  the REF allele for each POS. reuse the `--fasta-ref`
  option already used by `--tsv2vcf`
* updated the test and test files to use the new option
* add new option to the man page

Fixes #243


mcshane:
@jmarshall mentioned the `.fai` file in the test directory might cause problems on windows machines. We have a couple of other `fai` files already and not clear how to get test to run `samtools faidx`, so going to pull this in anyway.

PR #252, opened 2015-04-30
Allow to omit index with bcftools merge.

johanneskoester:
This pull request adds an option `--omit-index` to bcftools merge.
It allows to use bcftools merge with multiple named pipes, which can be very handy with large, sorted bcf files that shall not be stored on disk.

It needs a change in htslib, proposed with the following pull request:
https://github.com/samtools/htslib/pull/203

I don't know if my changes are appropriate and I am open to any suggestions.

### DIFF
--- a/vcfmerge.c
+++ b/vcfmerge.c
@@ -107,7 +107,7 @@ typedef struct
 {
     vcmp_t *vcmp;
     maux_t *maux;
-    int header_only, collapse, output_type, force_samples, merge_by_id;
+    int header_only, collapse, output_type, force_samples, merge_by_id, omit_index;
     char *header_fname, *output_fname, *regions_list, *info_rules, *file_list;
     info_rule_t *rules;
     int nrules;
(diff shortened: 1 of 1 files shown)


pd3:
Hi, 
thank you for the pull request.  In early versions of vcftools I was allowing to merge files without requiring an index. But it occasionally lead to unpleasant surprises and I grew less fond of the idea. Still, I can see that this can be sometimes useful, so maybe let's do it like this:

The bcftools side of things is OK, the merging still requires indexed files by default. However, also the default behaviour of the synced reader should stay unchanged. How about initializing the reader with an optional 'do_not_require_index_even_with_multiple_readers' flag? This could be done by passing a different value to the require_index flag, for example by introducing new defines

     sr->require_index = BSF_SR_REQUIRE_INDEX_THROW
     sr->require_index = BCF_SR_REQUIRE_INDEX_WARN

or maybe wrap it in a macro

      bcf_sr_require_index_throw(sr)
      bcf_sr_require_index_warn(sr)

or something like that.

johanneskoester:
Hi and thanks for considering the pull request.
Your idea sounds reasonable for me. I will provide an updated request ASAP.

mcshane:
Thanks. Closing this PR now. Although we are still open to a PR along the lines @pd3 describes above.

PR #258, opened 2015-05-11
Fix norm --site-win long option

noporpoise:
norm --site-win long option is incorrectly mapped to -W instead of -w. Since -W option does not exist, the --site-win long option does not work. This commit corrects this.


mcshane:
Thanks Isaac.

PR #264, opened 2015-05-15
Add per sample concordance correlation r to vcf stats

mp15:
Add per sample concordance correlation r to vcf stats.
Don't check if value is NULL before passing it to free(), free is guaranteed to work with NULL


mp15:
Urgh, test is failing because of edge case linux yields -nan and mac yields nan. Will investigate

jmarshall:
I'm sure you can think of something more user-friendly to print than `nan` :smile: — cf samtools/samtools@ca44680262884c9f14b09b525b3a55e37e583ec7 and others.

mp15:
Fixed up the bugs here. NA is now printed instead of nan, which should make this more useful for the plotting code it is intended for.

mp15:
NA value replaced with 0 as requested.

tommycarstensen:
Martin, just to be annoying... 0 is a terrible choice IMO for two reasons. Main reason: All plotting programs will interpret a zero as what it is. Minor reason: You could in theory have a correlation of 0 for a small set of variants; e.g. the 30 Sequenom SNPs.

R uses `NA` by default. Gnuplot has no default, but you can set it to be any value with `set datafile missing`:
http://gnuplot.sourceforge.net/docs_4.2/node172.html

Likewise numpy has `missing_values`:
http://docs.scipy.org/doc/numpy/reference/generated/numpy.genfromtxt.html

And finally Excel does `#N/A` by default.

mcshane:
This is merged as is, but @mp15 to open a new issue about writing out missing values. Writing out 0 is apparently used in multiple places already, so should update all for consistency rather than this one case.

mcshane:
issue opened as #272

PR #321, opened 2015-09-17
Update norm test VCF headers to v4.2

jpiper:
These VCF files contain features that were only introduced in v4.2 of the VCF spec (e.g. `Number=R`) and should be updated accordingly.


mcshane:
Thanks.

PR #343, opened 2015-10-28
bcftools norm incorrect GT splitting

jpiper:
Hi All,

It appears to me there's a pretty serious flaw with GT flag splitting in the normalisation. This is what I'm seeing:

Splitting multiallelic sites such as

    chr1	113938929	.	T	A,C	52.00	LowGQX;HighDPFRatio	SNVSB=0.0;SNVHPOL=21	GT:GQ:GQX:DP:DPF:AD	1/2:7:5:5:8:0,4,1

with a GT of `1/2` with `bcftools norm -m- <in>` incorrectly yields two records with a GT `1/0` and `0/1`, which should really be be `1/.` and `./1`

    chr1	113938929	.	T	A	52	LowGQX;HighDPFRatio	SNVSB=0;SNVHPOL=21	GT:GQ:GQX:DP:DPF:AD	1/0:7:5:5:8:0,4
    chr1	113938929	.	T	C	52	LowGQX;HighDPFRatio	SNVSB=0;SNVHPOL=21	GT:GQ:GQX:DP:DPF:AD	0/1:7:5:5:8:0,1

In the integration tests, I also notice sites with a `2` getting split to a `0` and a `1` site...

    20	275	.	A	C,G	999	PASS	INDEL;AN=2;AC=0,2	GT:PL:DP:FGF:FGI:FGS:FSTR	2:0,0,0:0:1e+06,2e+06,3e+06:1111,2222,3333:A,BB,CCC:WORD	2:0,0,0:0:1e+06,2e+06,3e+06:1111,2222,3333:A,BB,CCC:WORD

to

    20	275	.	A	C	999	PASS	INDEL;AN=2;AC=0	GT:PL:DP:FGF:FGI:FGS:FSTR	0:0,0:0:1e+06,2e+06:1111,2222:A,BB:WORD	0:0,0:0:1e+06,2e+06:1111,2222:A,BB:WORD
    20	275	.	A	G	999	PASS	INDEL;AN=2;AC=2	GT:PL:DP:FGF:FGI:FGS:FSTR	1:0,0:0:1e+06,3e+06:1111,3333:A,CCC:WORD	1:0,0:0:1e+06,3e+06:1111,3333:A,CCC:WORD

which should surely be `.` and `1` (For humans, you wouldn't expect a haploid call in this specific region - the spec says we would only expect this in M or Y for humans)

I have patched [here](https://github.com/samtools/bcftools/blob/develop/vcfnorm.c#L520) to replace

    gt[j] = bcf_gt_unphased(0) | bcf_gt_is_phased(gt[j]); // set to REF

with

    gt[j] = bcf_gt_missing | bcf_gt_is_phased(gt[j]); // set to REF

and altered the expected output from the unit tests accordingly, in case you find this is actually a bug rather than intended behaviour.


jpiper:
Dirty pull request - issued from wrong branch! closing. will reopen.

pd3:
Hi, thanks for this! 
It's actually not a bug but the intended behaviour. The original purpose of the -m option was to create a workaround for programs which cannot handle multiallelic sites (namely pbwt) - one can split the multiallelic sites, do the analysis, and then merge back. However, since people started using it in a more general way, we will think of ways how to make it more versatile and more in line with users' expectations.

jpiper:
Hey! Thanks for the super quick reply. I was just emulating how vt normalize handles these sites. I'll still put the clean pull request in for you in case you find it helpful :)

PR #348, opened 2015-11-10
add --threads option for output compression threads

mcshane:
* `--threads` option added to commands that output VCF/BCF
* add this new option to the docs


wkretzsch:
Sweet! Goodbye pigz, hello bcftools :D

PR #352, opened 2015-11-12
plot-vcfstats: plotting of sample names fixed

dlaehnemann:
Plotting of sample names instead of sample ID numbers (flags -s/--sample-names) in plot-vcfstats did not work as the first input to plt.xticks in <prefix>-plot.py was an array of strings instead of an array of integers for all by-sample plots. A simple call of int() around the row[0] argument in the list comprehension fixes that.

### DIFF
--- a/plot-vcfstats
+++ b/plot-vcfstats
@@ -877,7 +877,7 @@ sub plot_per_sample_stats
             \\tax1.set_ylabel('Ts/Tv')
             \\tax1.set_ylim(min(float(row[1]) for row in dat)-0.1,max(float(row[1]) for row in dat)+0.1)
             \\tif sample_names:
-            \\t\\t     plt.xticks([row[0] for row in dat],[row[7] for row in dat],**sample_font)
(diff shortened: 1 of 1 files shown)


pd3:
Could this be dependent on matplotlib version? I am sure it worked for others. Anyway, thanks for the patch!

dlaehnemann:
From what I can see, the underlying matplotlib-function set_xticks has always asked for a list of floats. And I wouldn't think that python does implicit type conversions (string to float) in any versions?!
While in this case, these numbers should always be integers, a more flexible solution might also be to use float() instead of int().

pd3:
Ah, you are right. This was fixed a while ago but was not integrated in the main repo yet https://github.com/pd3/bcftools/commit/e946ef864142c199fc11ae060f4272b9da37e208

dlaehnemann:
Sorry, didn't see that one. Feel free to close this one, then. Or change it to float(), or whatever you think is best.

pd3:
My fault, I should be quicker in getting bugfixes into the main repo. Sorry for the inconvenience and thanks again for looking into this.

mcshane:
Thanks.

PR #359, opened 2015-12-07
New concat --naive option for fast operations on large BCFs

mcshane:


vcfconcat.c
```diff
@@ -549,6 +553,114 @@ static void concat(args_t *args)
     }
 }
 
+BGZF *hts_get_bgzfp(htsFile *fp);
```
jmarshall: No.  Use the declaration in tbx.h, with a note as `#include <htslib/tbx.h> // for hts_get_bgzfp()`

vcfconcat.c
```diff
@@ -549,6 +553,114 @@ static void concat(args_t *args)
...
+static void naive_concat(args_t *args)
+{
+    // only compressed BCF atm
+
+    BGZF *bgzf_out;
+    if ( strcmp("-",args->output_fname) )
+        bgzf_out = bgzf_open(args->output_fname,"w");
+    else
+        bgzf_out = bgzf_dopen(fileno(stdout), "w");
```
jmarshall: No, unhelpful for Windows.  `bgzf_open` understands `"-"`, so these five lines can be one.

vcfconcat.c
```diff
@@ -549,6 +553,114 @@ static void concat(args_t *args)
...
+
+    int page_size = getpagesize();
+    char *buf = (char*) valloc(page_size);
```
jmarshall: No, do not use these functions.  cf samtools/htslib@45c14cc37c65577a75d4043562f832cece38e13f.

vcfconcat.c
```diff
@@ -549,6 +553,114 @@ static void concat(args_t *args)
...
+        {
+            if ( bgzf_write(bgzf_out, fp->uncompressed_block+nskip, fp->block_length-nskip)<0 ) error("Error: %d\n",fp->errcode);
+        }
+        if ( bgzf_flush(bgzf_out)<0 ) error("Error: %d\n",bgzf_out->errcode);
+
+
+        // Stream the rest of the file as it is, without recompressing, but remove BGZF EOF blocks
+        ssize_t nread, ncached = 0, nwr;
+        const int neof = 28;
+        char cached[neof];
```
jmarshall: No need to use a C99 VLA here.

Really it would be better to use `bgzf_check_EOF` or the like rather than needing to know what a trailer block looks like here.

vcfconcat.c
```diff
@@ -31,6 +31,7 @@ THE SOFTWARE.  */
 #include <htslib/vcf.h>
 #include <htslib/synced_bcf_reader.h>
 #include <htslib/kseq.h>
+#include <htslib/bgzf.h>
```
jmarshall: Add to the _vcfconcat.o_  prerequisites in the makefile.

vcfconcat.c
jmarshall: Please fix also in _reheader.c_.  Sigh…

pd3: I don't like solution hardwiring 32K as in https://github.com/samtools/htslib/commit/45c14cc37c65577a75d4043562f832cece38e13f. If portability is a concern, then the recommended replacement of `getpagesize` is `sysconf(_SC_PAGESIZE)` from `unistd.h`. Similarly `valloc` can be replaced with `posix_memalign`.

jmarshall: Using a page-aligned buffer was important in 1983 when doing page-aligned reads from a file.  Here the reads are not page-aligned and it is no longer 1983.  So either evidence is needed showing some gain from `valloc`, or anything beyond `malloc(some_reasonable_constant)` is premature optimisation.  And 32K is IMHO the absolute minimum reasonable constant.

pd3: I had no doubts you'd come to justify anything horrid, even hard-wired machine-specific constants, as long as it was you who put it there :-D
It is true that last time I did the benchmarking of this was ~10 years ago (not in 1983), but it did make a significant difference. It is possible that code optimization, plenty of available memory and better machines make this practice obsolete, but 1) can you claim you have benchmarked this on every machine out there 2) what is the harm?

jkbonfield: The harm in using obscure and deprecated functions is code portability.  IIRC on Windows I couldn't use sysconf either in the Staden Package as it simply didn't exist on MSC.  Maybe it does now, but it demonstrates the point still.  Hard coded numbers are obscure, but sufficient if commented AND PORTABLE!  

Besides, the whole argument over valloc vs malloc is kind of irrelevant when it could just be ```char buf[BUF_SIZE]``` with a define.   (I'm happy with either malloc or fixed buffer btw, but not with valloc.)

There's also nothing wrong in using a fixed buffer size.  If you _really_ care over what it should be, then the right thing to do is to stat the file and look at the filesystem preferred block size ```st_blksize```.

Anyway, as it currently stands is sufficient and we need to get the release out of the door!

pd3: Well, I suggested to use the portable versions of the now deprecated functions, the discussion might have ended there!

pd3: Last comment from me to this thread: I learned the aligned pagesize trick long time ago from linux coreutils. This is a well tested approach and tt really does make a difference, depending on the situation.

PR #371, opened 2016-01-14
New tool to get genotype intersection counts across samples of a vcf file.

dlaehnemann:
This tool is meant to get the counts of all possible subsets of the samples in one vcf file and output them in banker's sequence order (http://www.thelowlyprogrammer.com/2010/04/indexing-and-enumerating-subsets-of.html). It basically replicates the output that vcf-compare used to give for plotting of Venn diagrams, but does a proper genotype comparison for haploid and diploid samples instead of just looking at the presence/absence of a position. The banker's sequence order output can then be used to efficiently generate the files necessary for plotting the subset counts, e.g. using circos (see https://github.com/dlaehnemann/bankers2).

Included in this pull request are:
* adapted Makefile and main.c
* vcfgtisec.c, containing the new tool
* test/view.gtisec*.out files providing the checked results of testing the tool with all possible combinations of flags used (the command line used is documented in the output, thus visible in the files)

The code has been tested intensively and memory leaks have been removed using valgrind checking.

### DIFF
--- a/plugins/GTisec.c
+++ b/plugins/GTisec.c
@@ -1,0 +1,465 @@
+/*  vcfgtisec.c -- collect genotype intersection counts of all possible subsets
+                   of the present samples and output in banker's sequence order
+                   (in this sequence, the number of contained samples increases
+                   monotonically, a property that is e.g. useful for
+                   programatically creating plotting files for the R package
+                   VennDiagram or the plotting tool circos from the counts, as
+                   in the command line tools bankers2VennDiagram and
+                   bankers2circos at htpps://github.com/dlaehnemann/bankers2)
+
+    Author: David Laehnemann <david.laehnemann@hhu.de>
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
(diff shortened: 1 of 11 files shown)


dlaehnemann:
Just checking, that this pull request doesn't get overlooked. If there is anything missing or unclear, I'd be glad to explain or amend the pull request. Or if you have general doubts whether this should go into bcftools, I'd be glad to discuss...

pd3:
Hi David,
thank you for your pull request, this is nice. A plugin might be a better container for this kind of functionality. A minor comment about the human readable output: how about putting the numbers in the first column, that might be easier to parse visually.

dlaehnemann:
Hi Petr,

sure, switching the columns in verbose and Human readable output is a good idea. This would also ease parsing of verbose output, keeping the count info in the same column.

As to the plugin: That was my initial idea, and I started coding it up as a plugin, but quickly hit the limits of the init(), process() and destroy() functions. I can't recall the specific point, where I had something that didn't fit into the framework, now, though. I could have another look at that, though, if you'd very much prefer a plugin.

Also, this tool could be extended to different kinds of subset counts, which would probably be more likely to warrant a full subtool...

Best,
david

pd3:
You are right, the plugin framework had some limitations. This was recently fixed. If the plugin implements the `run` handler, bcftools now immediately pass the control over to the plugin, so should be OK.

dlaehnemann:
Thus, I take it that you would prefer this as a plugin? ;)

pd3:
Err, yes, sorry for not being clear.

dlaehnemann:
Alright, will try that ASAP and bring it in here.

dlaehnemann:
It's a plugin now. If there's anything else you need, let me know.

mcshane:
Hi David. Thanks for your work on this. It looks like you have test input and output files here, but no addition to test/test.pl. Could you add your tests in please? There were some very recent pushes on develop to ensure that plugin tests get run on travis too, so if you could rebase and squash down your commits we can get this merged in once the tests are ok.

Also, in your test files you will need to grep out the version related lines as they will change from commit to commit (e.g. `bcftools (1.2-221-g7b7e6b8-dirty+htslib-1.3-2-g6a745fe)`)

dlaehnemann:
Almost done with this, but have the following problem:
test.pl does not only test the output of a plugin against the provided test output file after grepping out the version line ("| grep -v ^##bcftools_"), but also without the grepping and that obviously fails. I could either implement a plugin-specific equivalent of --no-version or implement an explicit testing function for this plugin in test.pl -- any preferences or other ideas?

mcshane:
Can you include in the `args`? Something like:
```
test_vcf_plugin($opts,in=>'GTisec',out=>'view.GTisec.Hv.out',cmd=>'+ GTisec',args=>'-- -Hv | grep -v "^# This file was produced by bcftools"');
```

dlaehnemann:
That was easy... ;)

dlaehnemann:
Still working on the rebasing, so please don't merge yet.

dlaehnemann:
OK, I think I have now finally understood rebasing and managed to get you a clean squash and rebase of my plugin code commits. If all tests pass, feel free to merge -- and thanks for your patience and feedback.

mcshane:
Thanks. Have pulled in as 2f4a2b232103bffe673c7eb7f9e2e0304fb55af6

PR #414, opened 2016-04-26
bcftools mpileup

mcshane:
Bring mpileup over to bcftools. The motivation being that the `vcf/bcf` generation by mpileup is tightly coupled to the calling process in `bcftools call`. Having these in separate tools/repos made mistakes with mismatched versions too likely and hampered development as release of new features often needed to happen in both samtools and bcftools simultaneously.

See the individual commit messages for more information.


mcshane:
This is the beginning of bringing @pd3 bcftools mpileup changes over. Have started from scratch by bringing over the mpileup related files from current samtools develop and adding changes on top of that (current state is essentially a redo of pd3/bcftools/d7a93f5621106c7d1d6adbcb0d6ac6586117ea2e and the non-gvcf part of pd3/bcftools@cf3219c3237dbd447bfa31be967f1605f531bfd6).

Before bringing in the new things it would be good if people see if everything looks okay.

Candidates for removal:

1. `-g/--BCF`, `-v/--VCF`, `-u/--uncompressed` could all be dropped. Argument for keeping for the time being would be that users could simply change `samtools` <-> `bcftools` and commands would still work although with a warning to start using the `-O type` option instead.
2. `-D`, `-V`, `-S` have been deprecated since samtools 1.0, so probably time to remove them.

Candidates for updates to default:

1. The default for `--excl-flags` is `[UNMAP,SECONDARY,QCFAIL,DUP]` at the moment. @pd3 has added `SUPPLEMENTARY` to that list in his fork. Arguments for and against that?
2. `--bam-list` long option. Bother to rename, since we now take CRAM too? Silently allow `--bam-list`, but advertise something else, like `--aln-list`
3. `-C` internally we always set this to `-C50`. No idea why.
4. @jkbonfield spent some time looking at the `-d` depth option in samtools. Any reason to change?
5. others?

There are basically 3 features on the @pd3 fork to be added in after this.

1. `--gvcf`: mpileup gVCF creation. If we remove the old `-g` option, that could be used as the short option for this. This will essentially be pd3/bcftools@cf3219c3237dbd447bfa31be967f1605f531bfd6 and the gvcf part of pd3/bcftools@ee8210d2f81af007c6b2ad47d699527c50d0a58e 
2. `--samples`: include only selected samples. If we remove the old `-s` option, that could be used as the short option for this (pd3/bcftools@cf5c354859cf7af542ba4cb936233a5ef8d2eb88). We talked about adding a feature to map readgroups to new sample names (samtools/samtools#324) on the fly. Keep in mind when bringing this over.
3. `--regions/--regions-file`: To allow index jumping to multiple regions. e.g. all the non-chromosomal contigs. `-r` option can be repurposed for `--region/--regions`, but `-R` that is used in other bcftools commands for `--regions-file` is occupied by `-R/--ignore-RG` at the moment. This feature is tied up with regidx additions in htslib. pd3/bcftools@cfd7cf9092ffddc1ec0441a3c463bc2ac7084bad

mpileup.c
```diff
@@ -55,43 +55,6 @@ static inline int printw(int c, FILE *fp)
...
-            if (c == '=' || seq_nt16_table[c] == seq_nt16_table[rb]) c = bam_is_rev(p->b)? ',' : '.';
-            else c = bam_is_rev(p->b)? tolower(c) : toupper(c);
-        } else {
-            if (c == '=') c = bam_is_rev(p->b)? ',' : '.';
-            else c = bam_is_rev(p->b)? tolower(c) : toupper(c);
-        }
-        putc(c, fp);
-    } else putc(p->is_refskip? (bam_is_rev(p->b)? '<' : '>') : '*', fp);
-    if (p->indel > 0) {
-        putc('+', fp); printw(p->indel, fp);
```
jmarshall: Can also remove `printw()`, which is only used here in `pileup_seq()`.

jkbonfield:
My thoughts.

In general, unless it can be kept 100% identical (which for good reasons it cannot) then we should make a clean break and feel free to change whatever we wish.  Our end goal should be a command within bcftools that is internally consistent with other bcftools commands.  It's preferable for this to happen in one step rather than have two releases with changes in.

The caveat here is it obviously needs careful and explicit documentation outlning what has changed and what the mapping from old to new (if existing) is.

- I have no issues with removal of -g and -v if there is a more canonical way to do this.  Ideally with the move to bcftools it should follow any conventions put in place by the rest of bcftools.  -u I'm less sure about as it's used in many places as a synonym for uncompressed data, but again I don't really know bcftools conventions so I'll leave that up to you/Petr.

- Deprecated items; drop them to free up the much over-used single letter alphabet.

- --excl-flags shouldn't include supplementary.  Supplementary reads can be primary alignments (unless marked as secondary) and therefore shouldn't be filtered.

- -s and -S for samples / sample-file seems common in other bcftools commands, so seems a good fit.   Maybe -r/-R should be for regions and change the existing -R to something else (or long only?).

Other things I don't really have an opinion about.

mcshane:
Thanks James. I agree with all that. One other change in the name of consistency within bcftools should probably be to drop the `-l, --positions FILE` option in favour of the `-t, --targets/-T, --targets-file` options. This would require a new short option for the `-t, --output-tags` option though.

jkbonfield:
On Wed, May 25, 2016 at 03:53:00AM -0700, Shane McCarthy wrote:
> Thanks James. I agree with all that. One other change in the name of consistency within bcftools should probably be to drop the `-l, --positions FILE` option in favour of the `-t, --targets/-T, --targets-file` options. This would require a new short option for the `-t, --output-tags` option though.

Rather confusingly this is -b BED-file in samtools depth and -l in
mpileup.  The help text is somewhat obfuscated too as:

    "skip unlisted positions (chr pos) or regions (BED)"

It took me some time to parse the double negative and lack of
bracketting.  My initial mental reading implied skip ... regions
(BED).

Maybe reword it as "Only output specified positions (chr pos) or regions (BED)".

James

-- 
James Bonfield (jkb@sanger.ac.uk) | Hora aderat briligi. Nunc et Slythia Tova
                                  | Plurima gyrabant gymbolitare vabo;
  A Staden Package developer:     | Et Borogovorum mimzebant undique formae,
https://sf.net/projects/staden/   | Momiferique omnes exgrabure Rathi. 


-- 
 The Wellcome Trust Sanger Institute is operated by Genome Research 
 Limited, a charity registered in England with number 1021457 and a 
 company registered in England with number 2742969, whose registered 
 office is 215 Euston Road, London, NW1 2BE.

mcshane:
OK. I've added three more commits. First one removes the easily deprecated options. The second introduces the `--gvcf` generation from @pd3's branch. The third introduces sample selection options.

The final sync-up for mpileup with the @pd3 branch is pd3/bcftools@cfd7cf9092ffddc1ec0441a3c463bc2ac7084bad, which is waiting on the decision on the htslib regidx changes.

In terms of further possible option rearrangements:

1. include the option to add the `^` prefix to exclude samples for the `-s` and `-S` options.
2. replace `-l` by `-t` and `-T` for non-indexed region extraction, including the option to add the `^` prefix to exclude regions.
3. the `-t,--output-tags` short option will need to be changed to avoid clashing with the `--targets` option. Perhaps `-a` is a good option - standing for annotation? Could also change the long option to `--annotate TAG` or similar.
4. change `--exclude-RG FILE` to something like `--reagroups FILE` to select RGs with exclusion coming via the `^` prefix again. Also allow for a second column to set sample names as for samtools/samtools#324. Sample name mapping could also be defined by a second column in the `--samples-file FILE`. This reading should be somewhat flexible as I could imagine adding other RG or SM specific annotation to these files, such as contamination estimate that could be used to adjust the likelihood calculations.

mcshane:
This branch is ready to be merged now. We should review docs and consider further changes once people have tried out the functionality on develop.

mcshane:
Rebased prior to merge.

PR #415, opened 2016-04-29
stats: changes in allele frequency stats

mcshane:
bcftools stats:

* new `--af-bins` option for user specified allele frequency binning
* new `--af-tag` option so users can bin AF based on an external
    allele frequency such as 1000 genomes
* calculate aggregate r-squared instead of average r-squared in
    allele frequency bins
* output AF as fraction instead of percentage

plot-vcfstats:

* remove deprecated repeat-consistency stats from the summary table
* ` --prefix` always interpreted as output directory, the previous
  behaviour was reported as annoying
* all file paths in `plot.py` script relative so that customization
  and replotting does not require the original directory structure


mcshane:
@mp15 where are you with reviewing this?

PR #416, opened 2016-05-03
annotate: add `-m` support for VCF, not just tab-delimited annot files

mcshane:


jmarshall:
Merged as b9e0105eb21432ea098b53e09e53a4eec9dd9244.

PR #437, opened 2016-06-13
index: add option to set the path for the index file

### LINKED ISSUE #373: bcftools index to allow specifying the path to the index file
Similar to the `samtools index $bam $index` command we should allow `bcftools index` to specify the path of the created index file. This should enable (as requested by @jrandall) piping like this:
```
bcftools view -Ob in.vcf | tee out.bcf | bcftools index - out.bcf.csi 
```

mcshane:
* removed filetype testing from `main()` as this prevents indexing streamed input files
* do not have to set `-` for stdin
* fail if reading from stdin and index path not set
* stats error messages cleaned up - VCF index can be CSI or TBI, so don't specify
* remove unused `fn_out` variable from stats
* requires samtools/htslib#330

Resolves #373

[NEWS]
* index: now has an option to set the path of the index file and can index a file from stdin
    `bcftools view -Ob in.vcf | tee in.bcf | bcftools index -o in.bcf.csi`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 # Makefile for bcftools, utilities for Variant Call Format VCF/BCF files.
 #
-#   Copyright (C) 2012-2015 Genome Research Ltd.
+#   Copyright (C) 2012-2016 Genome Research Ltd.
 #
 #   Author: Petr Danecek <pd3@sanger.ac.uk>
 #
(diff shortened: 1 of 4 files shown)


mcshane:
@daviesrob or @jmarshall would either of you like to take a look at this before merging?

daviesrob:
Looks find to me, although it would be nice to get rid of the use of `alloca` in `main_vcfindex`.  `alloca` is not completely portable, and doesn't behave well if it overflows the stack (see http://stackoverflow.com/questions/1018853/why-is-the-use-of-alloca-not-considered-good-practice).  It might be better to `ksprintf` into a `kstring_t` to build the new name.

mcshane:
Thanks @daviesrob. I've updated and rebased the PR.

PR #441, opened 2016-06-20
gtcheck: improved cross-check mode

mcshane:
This new version is more robust to the variations in the number of compared
sites and reports error rate lines (ERR), the new default way of estimating
sample similarity. The program also attempts to do a simple hierarchical
clustering. If the new cluster does not work as expect (the real error rate
is different from the defaults), the `-c <min,max>` option can be tuned:

- the first "min" argument controls the typical error rate in lanelets
from the same sample. Lanelets with error rate smaller than this will be
always considered as coming from the same sample.

- the second "max" argument is the reverse: lanelets with error rate
bigger than the absolute value of this parameter will be always
considered as different samples. When the value is negative, the cutoff
may be heuristically lowered by the clustering engine. If positive, the
value is interpreted as a fixed cutoff.

Also the -g mode should be improved at some point.

- [x] add documentation to man page

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,7 +37,6 @@ TABIX  = $(HTSDIR)/tabix
 
 CC       = gcc
 CPPFLAGS =
-CFLAGS   = -g -Wall -Wc++-compat -O2 -Wpadded
 CFLAGS   = -g -Wall -Wc++-compat -O2
 LDFLAGS  =
 LIBS     =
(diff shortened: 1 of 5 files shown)


mcshane:
Need to add documentation to the man page.

mcshane:
@dkj Do you have someone who could take a look at this? I guess Carol has already tested from Petr's fork, but would be good to confirm changes in this PR.

PR #466, opened 2016-07-25
Make install should depend on $(PLUGINS)

kdmurray91:
Hi all,

Attached is a trival patch to ensure that the `install` makefile rule has a dependency on compiled plugins. Otherwise one cannot 'make install' in a single step.

Cheers,
Kevin


mcshane:
Thanks.

PR #479, opened 2016-09-07
vcf2bed plugin added

gtamazian:
The **vcf2bed** plugin converts a VCF file to the BED format. The obtained BED file contains four columns:

1. the sequence name;
2. the starting position of a variant;
3. the ending position of a variant that is equal to its starting position plus the REF allele length;
4. the variant ID.

### DIFF
--- a/plugins/vcf2bed.c
+++ b/plugins/vcf2bed.c
@@ -1,0 +1,77 @@
+/* plugins/vcf2bed.c -- convert a VCF file to the BED format. 
+
+   Copyright (c) 2016 
+
+   Author: Gaik Tamazian <mail (at) gtamazian (dot) com>
+
+   The MIT License
+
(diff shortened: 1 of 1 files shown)


pd3:
Hi,
thanks you for the pull request, looks good.

This functionality would go very well with the existing `query` command. How about we instead add `POS0`, `END` (and possibly also `END0`, for symmetry), then the `vcf2bed` output can be created with 

    bcftools query -f'%CHROM\t%POS0\t%END\t%ID\n'

and no additional plugin is needed?

gtamazian:
Hi Petr,

Thanks for pointing out the `bcftools query` tool. I agree that it would be better to add `POS0` and `END` to `query` than to introduce a separate plugin.

Although the BED format does not use inclusive end coordinates, `END0` may come in handy for some applications, so we should also add it.

PR #493, opened 2016-10-05
Update bcftools.txt

tommycarstensen:


pd3:
The documentation is correct, the whole list of FORMAT tags is negated in that example by the single caret character `^`. Although it may not seem so at first sight, it is unambiguous.

PR #503, opened 2016-10-24
Subset specific TAGS

mh11:
These changes allow to customise the naming of tags and calculation of alleles by providing
- a custom **prefixes** for **tags**
- a list of sample IDs to calculate the **tags** for
## Combined usage (prefix + samples)
By using these options in combination allows the calculation of e.g. **EUR**, **SAS** and *\* EAS*\* samples on the fly.

```
bcftools +fill-tags my.vcf.gz -- \
-S EUR_,my_eur.txt -S NOT_EUR_,^my_eur.txt \
-S SAS_,my_sas.txt \
-S EAS_,my_eas.txt | \
bgzip > my_annotated.vcf.gz
```
## Prefix

Using a prefix **ABC_** all tags (incl. description) will have the prefix added.
e.g. ABC_AN=.., ABC_AC=..,

## Samples
Using a sample list allows the calculation of **TAGS** specific to this group of samples.
e.g. -S my_list.txt


pd3:
Hi, thank you for this pull request. There is a [related plugin](https://github.com/pd3/bcftools/blob/develop/plugins/aggregate.c) which should be merged with fill-tags at some point. Could you perhaps modify your pull request according to that? It is more general in that it annotates multiple populations simultaneously.

mh11:
Thanks for the suggestion - your related plugin is very similar in this sense. I’ll give it a go to modify the pull request.

mh11:
The latest commit gives you the flexibility to use multiple ID list in different ways (include/exclude) to annotate vcf files with custom headers.

pd3:
Thank you for this. Can you please check the failing tests?

mh11:
Done - checks are updated, additional test added and all passing.

pd3:
Hi, just added a minor improvement in -S initialization (https://github.com/samtools/bcftools/commit/1322e047f9c94def0a650a8617803d9fbdf6258a) and merged. This pull request can be now closed. Thanks for this!

PR #525, opened 2016-12-08
Updated #stats section for the two-VCF-input case

tommycarstensen:
It's not specified anywhere in the documentation, that bcftools can calculate correlation and dosage correlation. I only assumed this, because Martin once asked me for the machine formula to calculate correlation in one loop. I don't think other people know. Maybe I'm the only ignorant one.


pd3:
Thank you for this documentation update!

Issue #534, opened 2017-01-19
"Cannot constrain to A" error handling


koelling:
I am trying to genotype a sample against a reference VCF file, including homozygous 0/0 calls.

To do this, I first generated a TARGETS file from the reference VCF file which contains the three columns "CHR\tPOS\tALLELES", with ALLELES = "REF,ALT". I then use `mpileup` with the `--positions TARGETS` argument to generate a pileup at all sites, followed by a `call` with `--keep-alts --multiallelic-caller --constrain alleles --targets-file TARGETS` to calculate genotype likelihoods for each site.

This seems to work in principle, however it simply crashes in the rare case that the ALT allele in the sample does not match the ALT allele in the reference data set. The error message I get in these cases is "Cannot constrain to A" (where A is the ALT allele).

Looking at the code, the error is thrown by this line: https://github.com/samtools/bcftools/blob/3f13cfee86739a8904db15d69e12607600fd3f47/mcall.c#L1296

This makes it essentially impossible for me to use `bcftools` because there is no proper way to avoid this error. The only solution I could think of at the moment would be to call genotypes first without constraining alleles, annotate each site with the expected alleles, compare the ALTs to the expected ALTs, find the sites where there is a mismatch, remove those from the original pileup and then rerun the caller, this time including the `--constrain` parameter. However, this would obviously be an incredibly complicated way of dealing with this.

Is there any way you could change the way mismatches between the expected and actual ALT alleles are handled?

I think it would be much better if `bcftools` simply printed a warning to STDERR and then ignored the site, instead of immediately exiting as soon as it encounters such a problem.

In theory I think this might be as simple as making these three changes:
- Change `mcall_constrain_alleles` to return `bool`
- Replace the call to `error` in `mcall_constrain_alleles` (linked above) by `return false;`
- Change `mcall_constrain_alleles(call, rec, &unseen)` to `if(!mcall_constrain_alleles(call, rec, &unseen)) { fprintf( stderr, "Skipping call because constraining alleles failed at %s:%d\n", bcf_seqname(call->hdr,rec),rec->pos+1); return 0; }`

However, it would be great if someone more familiar with the code could have a look at this!

pd3:
Now addressed by this commit https://github.com/samtools/bcftools/commit/6cb790c5a575a35bf0e60e9eb95f6ecbbeb2bb56.

Thanks for this and sorry for the delay

PR #535, opened 2017-01-19
when constraining alleles, skip calls with problems instead of crashing

koelling:
My suggestion for fixing #534 

Seems to work fine and solves my issue, but it would be great if someone could check to make sure there are no unintended consequences!

### DIFF
--- a/mcall.c
+++ b/mcall.c
@@ -1270,7 +1270,7 @@ void mcall_trim_numberR(call_t *call, bcf1_t *rec, int nals, int nout_als, int o
 // NB: in this function we temporarily use calls->als_map for a different
 // purpose to store mapping from new (target) alleles to original alleles.
 //
-static void mcall_constrain_alleles(call_t *call, bcf1_t *rec, int *unseen)
+static int mcall_constrain_alleles(call_t *call, bcf1_t *rec, int *unseen)
 {
     bcf_sr_regions_t *tgt = call->srs->targets;
     if ( tgt->nals>5 ) error("Maximum accepted number of alleles is 5, got %d\n", tgt->nals);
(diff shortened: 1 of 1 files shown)


pd3:
Hi, thank you for looking at this and for the pull request. We can accept it as a temporary workaround, but could you please make the function return 0 on success and non-zero on error? (I know the code is not consistent in this.) 
We should leave the issue #534 open, ideally the code should not be skipping such records.

koelling:
Okay, I have changed this now to return 0 on success and 1 on failure. I do think this is a bit confusing though, especially since mcall() seems to be returning 0 on failure. Maybe it would be better to include `stdbool.h` and make this return boolean `true` on success and `false` on failure?

pd3:
`mcall()` returns the number of used alleles. So far it was not handling errors and explicitly asking a site to be dropped. Perhaps a negative value should be used instead in such case

koelling:
I copied the approach of returning 0 on error from a bit further below in the function, where you also seem to return 0 if the site is not "callable": https://github.com/koelling/bcftools/blob/313f8b240b00580751db47965154b55e79b3463a/mcall.c#L1479

Happy to change it to -1, but I assume this would require further changes downstream.

pd3:
Yes you are right, sorry. This will require also a change in vcfcall.c, this line will have to be changed from `error` to `continue`.
https://github.com/samtools/bcftools/blob/develop/vcfcall.c#L852

koelling:
Hmm, wouldn't it be better to leave it as `return 0` to keep it consistent with the rest of the mcall() function? There are several other points where you return 0 to skip and move on to the next variant (e.g. when there are too many alleles, or if the CALL_VARONLY flag is set and is_variant is false). Or would you like me to change all of these to -1?

pd3:
If it is changed, it would need to be done properly throughout the function. But if you don't feel like it, you can leave it to us. Thanks

koelling:
No worries, I am happy to make the changes! One more question though: In vcfcall.c you do some additional handling of the return value depending on whether you are in gvcf mode or not:
https://github.com/samtools/bcftools/blob/develop/vcfcall.c#L852

In particular, if mcall returns zero you might still write it to the file if gvcf mode is on. So if I simply changed mcall() to always `return -1` on error and add a `if(ret == -1) continue;`, to vcfcall.c, this would change the output for gvcfs.

So maybe mcall() should still return 0 for sites that are okay in principle but just not a variant (nAC is zero or out_als is 1), while it should return -1 if the site falls for some reason (too many alleles/cannot constrain)?

Let me know if you agree and I will make the changes.

PR #546, opened 2017-02-03
mpileup: require the reference fasta by default

mcshane:
Piping `mpileup` into `call` using default options will produce
an empty VCF because REF is filled with `N` when the reference
is not present and `call` ignores such sites by default.
Seeing as bcftools/mpileup is only doing VCF output rather than
the pileup text output, we now require the reference by default.
Requirement can be switched off with the `--no-reference` option.

Resolves samtools/samtools#256


mcshane:
merged as 09fd67c2b0e9a8653d19d66c568ed27df8c2011b

PR #547, opened 2017-02-03
convert: use the new kputd from htslib

mcshane:
@pd3 does it make any sense to also use this for the `%f` output for prob3? Would change the format of some output.


mcshane:
merged as 6a5254b1bcec9869b681d68095c124a21fd8f387

PR #550, opened 2017-02-13
Remove test of tabix on test/large_chrom_tbi_limit.vcf

daviesrob:
This test fails when htslib is improved so that it reports an error when
trying to add an index entry beyond the highest position that the index
can handle.  Remove the test from here and put it into htslib itself
so that bcftools' `make test` carries on working after htslib is updated.

This update needs to go into bcftools before the related changes are pushed to htslib to avoid complaints about bcftools failing to build.


pd3:
Thank you for this

PR #556, opened 2017-02-23
cleanup clang compiler warnings

mcshane:
cleanup `tautological-constant-out-of-range-compare` and `unused function` warnings.


pd3:
Thank you

PR #564, opened 2017-03-02
Use htslib_static.mk to get static link flags

daviesrob:
Companion for https://github.com/samtools/htslib/pull/463 to make bcftools link the right libraries for various configurations of htslib.


mcshane:
merged as c3de3c4ecfccef064b9c6ea247349adf82d00214

Issue #388, opened 2016-03-08
-rdynamic problem for Mac OS 10.10.5


top1214:
```
$ make install
gcc -rdynamic  -o bcftools main.o vcfindex.o tabix.o vcfstats.o vcfisec.o vcfmerge.o vcfquery.o vcffilter.o filter.o vcfsom.o vcfnorm.o vcfgtcheck.o vcfview.o vcfannotate.o vcfroh.o vcfconcat.o vcfcall.o mcall.o vcmp.o gvcf.o reheader.o convert.o vcfconvert.o tsv2vcf.o vcfcnv.o HMM.o vcfplugin.o consensus.o ploidy.o version.o ccall.o em.o prob1.o kmin.o  htslib-1.3/libhts.a -lpthread -lz -lm -ldl  
gcc: error: unrecognized command line option ‘-rdynamic’
make: *** [bcftools] Error 1
```

I tried removing that flag from the makefile, and this happened

```
$ make install
gcc   -o bcftools main.o vcfindex.o tabix.o vcfstats.o vcfisec.o vcfmerge.o vcfquery.o vcffilter.o filter.o vcfsom.o vcfnorm.o vcfgtcheck.o vcfview.o vcfannotate.o vcfroh.o vcfconcat.o vcfcall.o mcall.o vcmp.o gvcf.o reheader.o convert.o vcfconvert.o tsv2vcf.o vcfcnv.o HMM.o vcfplugin.o consensus.o ploidy.o version.o ccall.o em.o prob1.o kmin.o  htslib-1.3/libhts.a -lpthread -lz -lm -ldl  
mkdir -p -m 755 /usr/local/bin /usr/local/share/man/man1 /usr/local/libexec/bcftools
mkdir: /usr/local/bin: Operation not permitted
mkdir: /usr/local/share/man/man1: Operation not permitted
mkdir: /usr/local/libexec/bcftools: Permission denied
make: *** [install] Error 1



```

top1214:
OK, i manually made /usr/local/libexec/bcftools, then removed from the makefile
`	$(INSTALL_DIR) $(DESTDIR)$(bindir) $(DESTDIR)$(man1dir) $(DESTDIR)$(plugindir)`
and I guess it worked
```
$ make install
install -p bcftools plot-vcfstats vcfutils.pl plugins/color-chrs.pl /usr/local/bin
Installing BAMToGASV and GASV...
Buildfile: build.xml does not exist!
Build failed
Creating pipeline scripts...
[...]

PR #576, opened 2017-03-24
bring over autoconf from samtools

### LINKED ISSUE #481: bcftools 1.3.1 hardcoded to build against htslib 1.3.1
htslib 1.3.2 has been released, but the bcftools 1.3.1 release is not set up to build against it. 

In my environment at Sanger, I was able to build like so:
`
touch /software/hgi/pkglocal/htslib-1.3.2/include/htslib.mk && make HTSDIR=/software/hgi/pkglocal/htslib-1.3.2/include HTSLIB=/software/hgi/pkglocal/htslib-1.3.2/lib/libhts.so.1.3.2 BGZIP=/software/hgi/pkglocal/htslib-1.3.2/bin/bgzip TABIX=/software/hgi/pkglocal/htslib-1.3.2/bin/tabix USE_GPL=1 GSL_LIBS="$(PKG_CONFIG_PATH=/software/hgi/pkglocal/gsl-1.16/lib/pkgconfig pkg-config --libs gsl) -Wl,-rpath=/software/hgi/pkglocal/gsl-1.16/lib" EXTRA_CPPFLAGS="-I. -I/software/hgi/pkglocal/htslib-1.3.2/include -DPLUGINPATH=\\\"/software/hgi/pkglocal/bcftools-1.3.1-htslib-1.3.2/libexec/bcftools\\\"" LDFLAGS="-L/software/hgi/pkglocal/xz-5.2.2/lib -L/software/hgi/pkglocal/bzip2-1.0.6/lib -L/software/hgi/pkglocal/zlib-1.2.8/lib -Wl,-rpath=/software/hgi/pkglocal/xz-5.2.2/lib -Wl,-rpath=/software/hgi/pkglocal/bzip2-1.0.6/lib -Wl,-rpath=/software/hgi/pkglocal/zlib-1.2.8/lib -Wl,-rpath=/software/hgi/pkglocal/htslib-1.3.2/lib" prefix=/software/hgi/pkglocal/bcftools-1.3.1-htslib-1.3.2 install
`

As intuitively obvious as that is, I feel like `bcftools` could benefit from an `autoconf` treatment so that one can just re-run more or less the same `./configure` line used for a previous build.

### LINKED ISSUE #573: Problem building with an installed htslib
The Makefile has a line:
include $(HTSDIR)/htslib_static.mk

If HTSDIR points to an installed htslib (rather than somewhere you have built it) then htslib_static.mk does not exist.
An installation directory is set to read-only, which means that when the makefile tries to make this file, it fails.

mcshane:
Still todo:

* `--enable-gsl` / `--with-gsl`
* `--disable-plugins` and make sure disambiguated from htslib plugins

Plus there may be a bunch of stuff I am missing...

Closes #573, closes #481


daviesrob:
I've just added my additions to this PR.  They add configure options for plugins and libgsl, and rearrange Makefile slightly so that variables get set at the correct times.  I've also updated INSTALL and made travis test both configure and non-configure builds.

mcshane:
Thanks alot @daviesrob! I've rebased against develop and squashed a few of the early commits.

I get the following warnings when running autoheader and autoconf. Anything to worry about?

```
configure.ac:81: warning: AC_CONFIG_SUBDIRS: you should use literals
../../lib/autoconf/status.m4:1099: AC_CONFIG_SUBDIRS is expanded from...
m4/ax_with_htslib.m4:55: AX_WITH_HTSLIB is expanded from...
configure.ac:81: the top level
```

mcshane:
Also, @daviesrob, the stuff with `-Wl,-export_dynamic` was brought in with b8d7e3bba79c170f217970d714da83486fe053be (see also link to samtools-help message in that commit) to solve #388. Would bringing back `-rdynamic` bring that issue up again?

daviesrob:
The warnings are due to AX_WITH_HTSLIB passing a variable ($HTSDIR) to AC_CONFIG_SUBDIRS.  It means bcftools' `./configure --help` won't print help messages for an embedded HTSlib's configure.  Otherwise it's nothing to worry about.

On `-rdynamic`, as far as I can make out it's actually slightly better supported than -Wl,-export_dynamic.  Our Mac mini claims:
`gcc version 4.2.1 (Based on Apple Inc. build 5658) (LLVM build 2336.11.00)`
`Apple clang version 4.0 (tags/Apple/clang-421.0.60) (based on LLVM 3.1svn)`
`@(#)PROGRAM:ld  PROJECT:ld64-133.3`
`ld` doesn't accept `-Wl,-export_dynamic`.  `gcc` is fine with `-rdynamic` while `clang` prints a warning about the option being ignored.

On newer installations the linker knows about `-Wl,-export_dynamic` and both `gcc` and `clang` will pass on the option if started with `-rdynamic`.  Given this, I think using `-rdynamic` is currently the least bad option.  It's also what htslib does if you try to build that with plugins.

It might be worth adding a test to configure to see if `-rdynamic` works, but the only platform I've found so far where it doesn't is illumos, and that has other issues too.

jmarshall:
You can configure with `./configure -Wno-syntax` to silence that warning — see the commit message and _README.md_ change in samtools/samtools@30bc5e2dbcbc13ca772a607f5a2fe1cfe12be709.

pd3:
Great stuff, thanks for this!

PR #587, opened 2017-04-05
Tidy upsfor C99 / _XOPEN_SOURCE=600 compliance.

jkbonfield:
Changes relate to:
- Replacement of alloca with malloc.
- Replacement of index with strchr.
- Use of \_\_func\_\_ if \_\_FUNCTION\_\_ is undefined.
- Inclusion of strings.h for strcasecmp and friends.
- Avoidance of arithmetic on void*.

Note we already use \_\_func\_\_ in some source files, so maybe we should
do a search and replace of the \_\_FUNCTION\_\_ uses instead of redefining
it?

Some non-standard components still exist, but are left as the solution
is possibly more painful than the problem (if indeed they ever do
become a problem).

Remaing warnings from gcc6 -g -Wall -O2 -pedantic -std=c99 -D\_XOPEN\_SOURCE=600

1) filter.c:1469:55: warning: format '%p' expects argument of type 'void \*', but argument 3 has type 'void (\*)(filter_t \*, bcf1_t \*, struct _token_t \*) {aka void (\*)(struct _filter_t \*, struct <anonymous> \*, struct _token_t \*)}' [-Wformat=]
         if ( tok->setter ) fprintf(stderr,"\t[setter %p]", tok->setter);
                                                       ^
This isn't something that is possible to resolve in an easy manner,
nor is it likely to be a portability concern.

2) vcfcnv.c:549:13: warning: string length '5195' is greater than the length '4095' ISO C99 compilers are required to support [-Woverlength-strings]
             "\n",
             ^~~~

This is trivial to fix, but we can fix it in the unlikely event it
ever becomes a genuine portability issue.

3) vcfplugin.c:265:20: warning: ISO C forbids conversion of object pointer to function pointer type [-Wpedantic]
     plugin->init = (dl_init_f) dlsym(plugin->handle, "init");

plugin->init is a dl_init_f type, typedef as int (z\*dl_init_f) (int, char \*\*, bcf_hdr_t \*, bcf_hdr_t \*);
Converting a function pointer to void\* invokes undefined behaviour,
but the very nature of dlsym relies on this and given it is defined in
POSIX then I believe this behaviour is also defined.

### DIFF
--- a/bam2bcf_indel.c
+++ b/bam2bcf_indel.c
@@ -370,10 +370,12 @@ int bcf_call_gap_prep(int n, int *n_plp, bam_pileup1_t **plp, int pos, bcf_calla
     }
     free(ref2); free(query);
     { // compute indelQ
-        int *sc, tmp, *sumq;
-        sc   = (int*) alloca(n_types * sizeof(int));
-        sumq = (int*) alloca(n_types * sizeof(int));
-        memset(sumq, 0, sizeof(int) * n_types);
+        int sc_a[16], sumq_a[16];
+        int tmp, *sc = sc_a, *sumq = sumq_a;
+        if (n_types > 1) {
+            sc   = malloc(n_types * sizeof(int));
+            sumq = calloc(n_types,  sizeof(int));
+        }
         for (s = K = 0; s < n; ++s) {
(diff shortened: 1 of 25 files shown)


pd3:
Thank you for this!

PR #606, opened 2017-05-04
Properly flush chain info

krooijers:
Hi,

I found the chain file was missing entries for some chroms, however inconsistently, as it depended on the ordering of the FASTA (and VCF/BCF) file. 
Some debugging led me to this fix.

best,
Koos


pd3:
Thank you for this!

PR #620, opened 2017-05-24
Hap/Legend vs Haps/Sample

jessehoff:
It is rather confusing but there is a distinction between haps/sample and hap/legend/sample, wherein 'hap' only reference to hap/legend/sample and haps references haps/sample. 
See: https://mathgen.stats.ox.ac.uk/genetics_software/shapeit/shapeit.html#hapsample

The actual name of the argument is correct. But users will avoid some headaches if this is clearer in the documentation. I don't think I got anything wrong in the cleanup here but I have missed something.

### DIFF
--- a/doc/bcftools.html
+++ b/doc/bcftools.html
@@ -884,20 +884,20 @@ depth information, such as INFO/AD or FORMAT/AD. For that, consider using the
 <span class="strong"><strong>--vcf-ids</strong></span>
 </span></dt><dd>
     output VCF IDs instead of "CHROM:POS_REF_ALT" IDs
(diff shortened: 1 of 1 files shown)


jessehoff:
also, the name of the --hapsample command should probably change to --hapssample

pd3:
The `html` is generated from this template `doc/bcftools.txt`, therefore I put your fix in the right place via this commit https://github.com/samtools/bcftools/commit/7eb67ba94fc666199a34a68e7d24230419f8377b and will close the pull request now.
Thank you for this!

PR #627, opened 2017-06-12
Remove the limit of 32 alternate alleles,

### LINKED ISSUE #621: fill-tags fails for large number of ALTs
Annotating a merged multi-sample VCF file with fill-tags fails with:
```bcftools: plugins/fill-tags.c:404: process: Assertion `rec->n_allele < 8*sizeof(int)' failed. ```

The variant causing the issue has 33 ALTs (different INDELs) and the failure is due to an assertion of (I presume) a maximum of 32 ALTs. 

Not sure if this should be fixed or not, but I thought it's worth noting.

### PR DESCRIPTION
**Pull request by pd3**
use kbitset instead of int. Resolves #621 once/if the
pull request https://github.com/samtools/htslib/pull/555
is merged.

### CODE REVIEW

**Comment by jmarshall** [comment]
This limit could be fixed using the existing `kbitset_t` API, without needing the new `kbs_resize()` function proposed in the HTSlib PR.

In reality, each time `process()` is called, it `kbs_init()s` a new bitset of an appropriate size for the number of alleles in that `process()` invocation's record. There is conceptually a separate bitset for each invocation, but to avoid memory allocations for every variant record in the input file the code stores a single bitset object in `args` and reuses it each time.

This code could record the maximum n_allele itself and recreate the bitset when necessary using the existing `kbitset_t` functions:

```c
// instead of  if (kbs_resize(…))
if (rec->n_allele > args->max_n_allele) {
    kbs_destroy(args->bset);
    args->max_n_allele = rec->n_allele;
    args->bset = kbs_init(args->max_n_allele);
}
```

thus avoiding excess memory allocations just as the proposed `kbs_resize` would.

Issue #624, opened 2017-06-05
plot-vcfstats non-zero status 256


price333:
Hi,

I'm getting the following error running plot-vcfstats:

Plotting graphs: python plot.py
Traceback (most recent call last):
  File "plot.py", line 219, in <module>
    for i in range(n): col[i] = cm(1.*i/n)
TypeError: 'range' object does not support item assignment
The command exited with non-zero status 256:
        python plot.py

 at /data/user/bin/plot-vcfstats line 76.

Any help would be greatly appreciated!

pd3:
Interesting error. Would you mind sharing the generated plot.py script? There is an email address on my profile page.

Also what version of python and matplotlib is this?

            python --version
            python 
            >>> import matplotlib
            >>> matplotlib.__version__
            '1.4.3'

price333:
Thank you for your reply. So I tried again with Python version 2.7.13 (matplotlib 2.0.2+4197.geb6e422ff) and get the following error:

Plotting graphs: python plot.py
Traceback (most recent call last):
  File "plot.py", line 262, in <module>
    ax1.ticklabel_format(style='sci', scilimits=(0,0), axis='y')
  File "/data/borne/virtualenv/test/lib/python2.7/site-packages/matplotlib-2.0.2+4197.geb6e422ff-py2.7-linux-x86_64.egg/matplotlib/axes/_base.py", line 2610, in ticklabel_format
    "This method only works with the ScalarFormatter.")
AttributeError: This method only works with the ScalarFormatter.
The command exited with non-zero status 256:
        python plot.py

 at /data/user/bin/plot-vcfstats line 76.
        main::error("The command exited with non-zero status 256:\x{a}\x{9}python plot.py\x{a}\x{a}") called at /data/user/bin/plot-vcfstats line 276
        main::plot(HASH(0x22c0e70)) called at /data/user/bin/plot-vcfstats line 65


The previous error I posted was Python version 3.5.2 (matplotlib 2.0.2+4197.geb6e422ff). I'll email you both plot.py files now.

pd3:
Thank you for sharing the data. I can now confirm this is an issue with matplotlib differences between versions. 

Editing the script like this might help (I don't have 2.x installed anywhere so cannot try it myself at this moment):
http://matplotlib.1069221.n5.nabble.com/How-do-you-use-ticklabel-format-td11642.html

Alternatively, as a quick workaround you could comment out offending lines.

PR #645, opened 2017-07-10
Python3 fix for plot-vcfstats

nsoranzo:
Fix for the error:
```
TypeError: 'range' object does not support item assignment
```
in generated `plot.py`.

Fix https://github.com/samtools/bcftools/issues/624 .


pd3:
Thank you for this!

PR #652, opened 2017-07-23
allow downstream variables (e.g. plugindir) to evaluate libexecdir

nieder:
Using --libexecdir=/path/FOO in ./configure is currently useless because config.mk doesn't include it, and so things like plugindir (which defaults to $(libexecdir)/bcftools) uses the default /path/libexec. This makes sure that libexecdir is properly taken from the ./configure options.


pd3:
Thank you

PR #660, opened 2017-08-08
Fix mendelian plugin not recognizing some cases

oronnavon:
The "mendelian" plugin currently fails to recognize the following case as mendelian inconsistency: both the father and the mother carry two copies of the REF allele, but the child carries a copy of the ALT allele.  Issue was fixed, and test cases were added to cover this and similar cases.


oronnavon:
Hi @pd3 / @lh3, could you please take a look and merge if possible?

pd3:
Thank you for this! Just merged your changes via 81c6dd30, plus fixed the restriction of biallelic sites and added support for sex chromosomes.

PR #688, opened 2017-10-10
Fix for new matplotlib

### LINKED ISSUE #687: plot-vcfstats ScalarFormatter error under latest matplotlib (2.1.0)
Under the latest matplotlib release (2.1.0) when I run plot-vcfstats, while it is executing the generated plot.py code I get an error like this
```
AttributeError: This method only works with the ScalarFormatter
```
which happens during the method call `ax1.ticklabel_format(...)` in the block of code contained in the `if plot_hethom_by_sample:` statement (so it only happens when you call bcftools stats with the `--samples` option).

I'm no expert on matplotlib, but it looks like it's interpreting some of the points that are read in as categorical values instead of numbers (because they're formatted as strings) so the Formatter becomes a StrCategoryFormatter instead of a ScalarFormatter (probably related to one of the new features in this release, "support for string categorical values").

Most of the `*.dat` files are read in with the pattern `dat.append([float(x) for x in row])`, but `tstv_by_sample.*.dat` just has `dat.append(row)`, since it contains sample names which can't be converted to a float, which causes the numbers to be interpreted as strings.

I think this is something I actually might be able to do a fix for, so I'll try and submit a PR for it, but I figured I'd submit this first so I have an issue to reference in the PR. Sorry if I'm doing this all wrong, I'm pretty new to contributing to projects like these.

mrphyja:
Reads in some values as float instead of str, resolves #687. Also resolves a problem not mentioned in that issue where the depth plot didn't display right because numbers were being interpreted as categories.

### DIFF
--- a/misc/plot-vcfstats
+++ b/misc/plot-vcfstats
@@ -871,7 +871,7 @@ sub plot_per_sample_stats
             with open('$img.dat', 'r') as f:
             \\treader = csv.reader(f, 'tab')
             \\tfor row in reader:
-            \\t\\tif row[0][0] != '#': dat.append(row)
+            \\t\\tif row[0][0] != '#': dat.append([float(x) for x in row[:7]] + [row[7]])
 
             if plot_tstv_by_sample:
             \\tfig = plt.figure(figsize=(2*$$opts{img_width},$$opts{img_height}*0.7))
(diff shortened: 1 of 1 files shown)


pd3:
Thank you for this!

PR #700, opened 2017-10-29
getopt_long() returns int, not char

ginggs:
This fails on architectures where char is unsigned by default. Thanks @AdrianBunk.


pd3:
Thank you

PR #701, opened 2017-10-29
Prevent pointer overflow on 32-bit architectures

ginggs:


vcfnorm.c
```diff
@@ -94,7 +94,7 @@ static inline int replace_iupac_codes(char *seq, int nseq)
 }
 static inline int has_non_acgtn(char *seq, int nseq)
 {
-    char *end = nseq ? seq + nseq : seq + UINT32_MAX;   // arbitrary large number
+    char *end = nseq ? seq + nseq : seq + (UINT32_MAX >> 2);   // arbitrary large number
```
pd3: Thank you for looking at this. A better fix here would be to assume nseq>0 here and fix the callers to pass rec->shared.l instead of 0.

ginggs:
@pd3 here's my attempt at a better fix.  I couldn't find `rec->shared.l`, did you mean `line->shared.l`, or did I miss something?

pd3:
Sorry, it was `line` indeed. Thank you

PR #709, opened 2017-11-09
Typo in consensus example

martinjvickers:
I believe this is a typo and should have been `samtools` not `bcftools`.


pd3:
Hi, it is not a typo. The `mpileup` command has been transferred to `bcftools`. This is to make `mpileup` and `call` in the same binary to avoid version issues. But thanks for the pull request anyway.

martinjvickers:
Yeah I just realised after finding a another example hence why I closed the pull request.

PR #713, opened 2017-11-17
Replaced "Phred scaled" with "Phred-scaled"

eyherabh:
Replaced "Phred scaled" with "Phred-scaled" as the latter (with the hyphen) is used everywhere else in Samtools, and consistency makes parsing the files easier.


pd3:
Thank you. It's good to make it consistent. But note that parsers should not rely on the Description part of the header definitions.

eyherabh:
Thanks for the tip. Indeed, but sadly, in my limited experience, the description has so far been more reliable in indicating the scale of fields such as GT, GP or PL (log10, linear or phred) than the vcf standard specified in the file format. Nevertheless, the description is just one additional piece of information for verifying the consistency of the VCF file before processing it.

PR #769, opened 2018-04-06
fix not removing "@" from filenames in filtering expressions

pjagiello:
There is a silly bug introduced recently: the "@" from filtering expressions was not removed, so for example this fails:

> bcftools view --include ID!=@list.txt input.vcf

like that:

> Could not read: @list.txt                                                                                                                                                    

I compared the code to older versions, and I believe this was missed when changing path expansion functions in filter.c.  Fix is pretty simple.


pd3:
Indeed, thank you for the fix!

PR #792, opened 2018-05-15
Quoting in the test script.

jkbonfield:
The -M ? in bcftools consensus test gives the wrong answer if you have
a single character file in your current working directory (I had _).


pd3:
Thank you!

PR #821, opened 2018-06-27
Use bgzf_open() to read FASTA file

jmarshall:
Avoid `hts_open()` as that is documented as being for SAM/BAM/CRAM or VCF/BCF files.

See discussion in samtools/htslib#721 for context.


pd3:
Thank you

PR #830, opened 2018-07-04
Silence warnings and fix error messages

jmarshall:
Silence minor warnings with GCC 8.1.0.

### DIFF
--- a/vcfsom.c
+++ b/vcfsom.c
@@ -351,9 +351,9 @@ static som_t *som_init(args_t *args)
     som->bmu_th = args->bmu_th;
     som->size   = pow(som->nbin,som->ndim);
     som->w = (double*) malloc(sizeof(double)*som->size*som->kdim);
-    if ( !som->w ) error("Could not alloc %"PRIu64" bytes [nbin=%d ndim=%d kdim=%d]\n", sizeof(double)*som->size*som->kdim,som->nbin,som->ndim,som->kdim);
+    if ( !som->w ) error("Could not alloc %zu bytes [nbin=%d ndim=%d kdim=%d]\n", sizeof(double)*som->size*som->kdim,som->nbin,som->ndim,som->kdim);
     som->c = (double*) calloc(som->size,sizeof(double));
-    if ( !som->w ) error("Could not alloc %"PRIu64" bytes [nbin=%d ndim=%d]\n", sizeof(double)*som->size,som->nbin,som->ndim);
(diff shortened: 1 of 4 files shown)


vcfsom.c
```diff
@@ -351,9 +351,9 @@ static som_t *som_init(args_t *args)
     som->bmu_th = args->bmu_th;
     som->size   = pow(som->nbin,som->ndim);
     som->w = (double*) malloc(sizeof(double)*som->size*som->kdim);
-    if ( !som->w ) error("Could not alloc %"PRIu64" bytes [nbin=%d ndim=%d kdim=%d]\n", sizeof(double)*som->size*som->kdim,som->nbin,som->ndim,som->kdim);
+    if ( !som->w ) error("Could not alloc %zu bytes [nbin=%d ndim=%d kdim=%d]\n", sizeof(double)*som->size*som->kdim,som->nbin,som->ndim,som->kdim);
     som->c = (double*) calloc(som->size,sizeof(double));
-    if ( !som->w ) error("Could not alloc %"PRIu64" bytes [nbin=%d ndim=%d]\n", sizeof(double)*som->size,som->nbin,som->ndim);
+    if ( !som->w ) error("Could not alloc %zu bytes [nbin=%d ndim=%d]\n", sizeof(double)*som->size,som->nbin,som->ndim);
```
pd3: The `fprintf` man page warns about the `z` modifier `Linux libc5 has Z with this meaning.  Don't use it.`. 
The modifier `PRIu64` is used in quite a few other places in bcftools, what is wrong with it here?

pd3:
Thank you for the pull request. Fixed here https://github.com/samtools/bcftools/commit/4dc4b5daca708b0c4836f7141833741b82464c61 as suggested, just avoided the `%zu` modifier.

vcfsom.c
jmarshall: libc5 was twenty years ago. The man page is saying don't use libc5's `Z` (capital Z).

daviesrob: There's nothing wrong with the `z` modifier, it's part of C99.  It's also used in a few places in both samtools and htslib, for example:
```
bgzf.c: hts_log_error("Read block operation failed with error %d after %zd of %zu bytes", fp->errcode, bytes_read, length);
```

For future updates it would be best to use `z` when the parameter is a `size_t`.  Less obscure and less typing too.

pd3: OK, thank you for explaining.

pd3:
This has been incorporated in some way, closing this now. Thank you.

PR #835, opened 2018-07-11
Fixed buffer overrun in bcftools view -c

jkbonfield:
This is a security issue, albeit minor as the trigger for the problem
is in the command line arguments being too long rather than data
driven.

It's also a real bug though as longest supported string of "nonmajor"
is 9 bytes (including nul) rather than the 8 expected.


pd3:
Thank you, much appreciated

PR #836, opened 2018-07-11
Fix premature loop termination in bam_smpl_add_bam()

daviesrob:
If the sample name was mapped to a different value in bsmpl_keep_readgroup(), variable `r` could end up pointing outside the header, which made the comparison `p = q > r ? q : r;` invalid. Depending on where `r` ended up pointing, the loop through the header could terminate before all the @RG lines were processed. Fix this by finding out where the line ends and setting `p` to that instead.

Add a check to ensure that found @RG strings are either at the start of the header or immediately after a newline, to ensure what is being scanned really is a read group header.

Found when testing bcftools on alpine linux, which uses musl instead of glibc.


pd3:
Thank you, much appreciated

PR #837, opened 2018-07-12
Add package install suggestions to INSTALL

daviesrob:
@pd3 Please could you cast your eye over this and let me know if it looks better than the original text.


pd3:
This is great, thank you

PR #842, opened 2018-07-17
Fix tautological compare warnings from clang

daviesrob:
Looks like these may have caused missing values to be handled
incorrectly in some cases.

Thanks to @jmarshall for reporting.


pd3:
Thank you both @jmarshall and @daviesrob , for noticing and for fixing.

PR #844, opened 2018-07-19
Use utility function to clamp lookup table index

jmarshall:
Silences the Clang warnings added to _vcfroh.c_ in 26b28453ab5d27c7041e68b3f510f9818220433d.

It seems this function was broken by eaf74ca1625910c0a75d6c391e8e1aac9040af8a, and the recent commit restored the correct behaviour. This year of breakage was not detected by the test suite, so tests exercising this code should be added.

All versions of this code do something unfortunate if `p[irr]` is negative, which presumably could happen given a suitably malicious or fuzzed input. The `max255()` function could be extended to do something appropriate for negative `i` too.

### DIFF
--- a/vcfroh.c
+++ b/vcfroh.c
@@ -130,6 +130,11 @@ void *smalloc(size_t size)
     return mem;
 }
 
+static inline int max255(int i)
+{
+    return i < 256 ? i : 255;
+}
+
 static void init_data(args_t *args)
 {
     int i;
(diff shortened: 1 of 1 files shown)


pd3:
Thank you for this, good to see clang happy now. The case of negative index is handled a few lines above, so that should be fine.

jmarshall:
Re negative: oops, saw that yesterday but forgot about it today.

Tests would still be good.

PR #847, opened 2018-07-20
Add tectonic support to plot-vcfstats

nsoranzo:
Use [tectonic](https://tectonic-typesetting.github.io) if `pdflatex` is not found in `PATH`.

Do not use `.` in image filenames to prevent issues with `\includegraphics` for both `pdflatex` and `tectonic`, see tectonic-typesetting/tectonic#187 .

Also, consolidate `$$opts{rasterize}`, `$$opts{pdf_plots}` and `$$opts{fmt}` variables in `$$opts{img_fmt}`, which could support other formats in the future (e.g. eps).


pd3:
Thank you for this. I am happy to add support for tectonic. Could you please update the pull request to include the square brackets as mentioned in https://github.com/tectonic-typesetting/tectonic/issues/187?

nsoranzo:
@pd3 Yep, already working on it!

nsoranzo:
@pd3 I've rebased the second commit to not change filenames, but use your curly-brackets workaround. Thanks for the support!

PR #946, opened 2019-01-21
Fix warnings due to function attributes added in htslib

daviesrob:
HTSlib commit [fe9c2f8f](https://github.com/samtools/htslib/commit/fe9c2f8fb301b3fca3db204b4147c1d37f83d594) added HTS_RESULT_USED to bcf_hdr_sync(), bcf_hrec_add_key(), bcf_hrec_set_val() and hrec_add_idx().  The last three were also changed to return int instead of void.

This adds checks on the return values to fix the warnings in bcftools caused by the htslib update.

### DIFF
--- a/version.c
+++ b/version.c
@@ -25,6 +25,7 @@ DEALINGS IN THE SOFTWARE.  */
 #include <stdarg.h>
 #include <stdlib.h>
 #include <stdio.h>
+#include <errno.h>
 #include <htslib/hts.h>
 #include "bcftools.h"
 #include "version.h"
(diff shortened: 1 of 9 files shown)


daviesrob:
Yes, but I'm leaving that for future work.  I don't want to make this PR too complicated.

version.c
```diff
@@ -44,6 +45,21 @@ void error(const char *format, ...)
...
 }
 
+void error_errno(const char *format, ...)
+{
+    va_list ap;
+    va_start(ap, format);
+    vfprintf(stderr, format, ap);
+    va_end(ap);
+    if (errno) {
+        fprintf(stderr, " : %s\n", strerror(errno));
```
jmarshall: Need to sample `errno` before the `vfprintf` above, so might as well be the first thing in `error_errno()` before the `va_start` stuff too.

Perhaps no space before the colon (`": %s\n"`), depending on style in other htslib/bcftools error messages?

daviesrob: Good spot; should be fixed now.

pd3:
Looks great, thanks!

PR #954, opened 2019-02-09
Add missing filter command to example

dvolk:


pd3:
Thank you very much for this!

PR #964, opened 2019-02-19
Undefined behavior for bcftools annotate --keep-sites

dmckean:
As documented, `--keep-sites` for bcftools annotate should not accept an argument. Short option `-k` behaves as defined.


pd3:
Thank you for noticing the typo and for the fix.

Issue #983, opened 2019-03-18
Replace FORMAT expression function AVG with MEDIAN, MEAN, and MODE


dancooke:
From the [expressions documentation](
https://samtools.github.io/bcftools/bcftools.html#expressions):

    function on FORMAT tags (over samples) and INFO tags (over vector fields)

    MAX, MIN, AVG, SUM, STRLEN, ABS, COUNT

It's not clear if `AVG` is the mean, median, or mode. I think replacing `AVG` with `MEAN`, `MEDIAN`, and `MODE` would make this clearer and more useful.

pd3:
The function `AVG` is the arithmetic mean, this should be updated in the man page. 

Thank you for the pull request. Can you make the failing travis errors go away and add documentation to doc/bcftools.txt?

dancooke:
@pd3 Done.

pd3:
Thanks! We will need to add tests for this new functionality, are you happy to include them in the pull request as well?

Also please don't remove AVG completely from the documentation, as for most people "average" is synonymous with arithmetic mean. (I guess you are coming from a different background, so our opinions may differ on this.)

PR #985, opened 2019-03-19
Add MEAN, MEDIAN, MODE, STDEV format expression functions

dancooke:
Addresses #983, and also adds a `STDEV` function to compute the standard deviation.

The mode function may need reviewing. I implemented a version that discretises using integer-defined intervals in order to ensure that integer typed values will compute the mode as expected.

### DIFF
--- a/filter.c
+++ b/filter.c
@@ -1213,7 +1213,7 @@ static int func_median(filter_t *flt, bcf1_t *line, token_t *rtok, token_t **sta
         if ( !bcf_double_is_missing(tok->values[i]) ) { vals[n] = tok->values[i]; n++; }
     if ( n )
     {
-        rtok->values[0] = vals[0]; //median(vals, n);
+        rtok->values[0] = median(vals, n);
         rtok->nvalues   = 1;
     }
     free(vals);
--- a/filter.c
+++ b/filter.c
@@ -152,11 +152,15 @@ struct _filter_t
 #define TOK_PERLSUB 27
 #define TOK_BINOM   28
 #define TOK_PHRED   29
+#define TOK_MEAN    30
+#define TOK_MEDIAN  31
+#define TOK_MODE    32
+#define TOK_STDEV   33
 
-//                      0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19 20 21 22 23 24 25 26 27 28 29
-//                        ( ) [ < = > ] ! | &  +  -  *  /  M  m  a  A  O  ~  ^  S  .  l  f  c  p  b  P
-static int op_prec[] = {0,1,1,5,5,5,5,5,5,2,3, 6, 6, 7, 7, 8, 8, 8, 3, 2, 5, 5, 8, 8, 8, 8, 8, 8, 8, 8};
-#define TOKEN_STRING "x()[<=>]!|&+-*/MmaAO~^S.lfcp"
+//                      0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19 20 21 22 23 24 25 26 27 28 29 30 31 32 33
+//                        ( ) [ < = > ] ! | &  +  -  *  /  M  m  a  A  O  ~  ^  S  .  l  f  c  p  b  P  n  i  d s
+static int op_prec[] = {0,1,1,5,5,5,5,5,5,2,3, 6, 6, 7, 7, 8, 8, 8, 3, 2, 5, 5, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8};
+#define TOKEN_STRING "x()[<=>]!|&+-*/MmaAO~^S.lfcpmids"
 
 // Return negative values if it is a function with variable number of arguments
 static int filters_next_token(char **str, int *len)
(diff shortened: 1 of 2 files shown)


dancooke:
Note that I removed `AVG from the documentation as I would consider this function depreciated.

filter.c
```diff
@@ -1168,6 +1176,135 @@ static int func_avg(filter_t *flt, bcf1_t *line, token_t *rtok, token_t **stack,
...
+            // discretize real values
+            int *int_vals = (int*) malloc(n * sizeof(int));
+            int i;
+            for (i=0; i<n; i++) int_vals[i] = vals[i];
+            // find mode of discrete samples
+            int mode_count = 0, mode_idx = 0;
+            for (i=0; i<n; i++) {
+                int j;
+                for (j = i + 1; j < n && vals[i] == vals[j]; ++j);
+                int count = j - i;
```
valeriuo: You could speed up the algorithm by incrementing `i` with `count` at the end of the outer loop, so that you skip the values from the current group.

filter.c
```diff
@@ -1168,6 +1176,135 @@ static int func_avg(filter_t *flt, bcf1_t *line, token_t *rtok, token_t **stack,
...
+{
+    switch (n) {
+        case 0: return 0;
+        case 1: return vals[0];
+        case 2: return vals[0] < vals[1] ? vals[0] : vals[1];
+        default:
+        {
+            qsort(vals, n, sizeof(double), compare_doubles);
+            // discretize real values
+            int *int_vals = (int*) malloc(n * sizeof(int));
```
valeriuo: malloced addresses should be checked for nullness.

filter.c
```diff
@@ -1168,6 +1176,135 @@ static int func_avg(filter_t *flt, bcf1_t *line, token_t *rtok, token_t **stack,
...
+                    mode_idx = i;
+                    mode_count = count;
+                }
+            }
+            free(int_vals);
+            return median(vals + mode_idx, mode_count);
+        }
+    }
+}
+static int func_mode(filter_t *flt, bcf1_t *line, token_t *rtok, token_t **stack, int nstack)
```
valeriuo: `func_mode`, `func_median` and `func_stdev` are similar. It would be preferable to have a single method with an additional parameter in the form of a function pointer, which, in the actual call, will be replaced by `mode`, `median` or `stdev`.

filter.c
```diff
@@ -1168,6 +1176,135 @@ static int func_avg(filter_t *flt, bcf1_t *line, token_t *rtok, token_t **stack,
...
+            qsort(vals, n, sizeof(double), compare_doubles);
+            // discretize real values
+            int *int_vals = (int*) malloc(n * sizeof(int));
+            int i;
+            for (i=0; i<n; i++) int_vals[i] = vals[i];
+            // find mode of discrete samples
+            int mode_count = 0, mode_idx = 0;
+            for (i=0; i<n; i++) {
+                int j;
+                for (j = i + 1; j < n && vals[i] == vals[j]; ++j);
```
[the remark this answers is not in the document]
dancooke: Yes. Fixed in 2583a41dea06992dd33738fe5374ea86dc176817.

filter.c
dancooke: Addressed in d985011bb63955b7e1d6a8e8bda6427d41d30e91.

filter.c
dancooke: Addressed in 9d657a5edf2e71b3aa98fe9ac52ff17915af049e.

filter.c
dancooke: Checked how? I don't see any checks on other `malloc`s in filter.c.

dancooke:
Regarding @pd3 [comment](https://github.com/samtools/bcftools/issues/983#issuecomment-474874410) to keep `AVG` in the docs. Do you want this to be added back to the list of functions? I think this could cause confusion (why have both `AVG` and `MEAN` etc?).

filter.c
pd3: If `mode` is supposed to return the most frequent value, I suggest using `khash` instead, sorting values is not necessary and would not scale well when used on FORMAT fields in VCFs with many samples.

dancooke: I don't have any familiarity with the `k` library so I'm not able to make this change. Another option - if you don't want to use `qsort` - may be to use a radix sort or bucket sort. If the values are actually integers, which would probably be the most common case as the mode of real values is not very useful, then bucket sort would likely be the most efficient option (if the maximum absolute value is not too big).

pd3: For this problem most efficient is hashing, any type of sort will be slower. We could add it after merging the pull request.

However, I am not completely convinced about adding the `mode` function. It leads to complications for floats, the outcome depends on discretization details. Do you have a practical use case in mind?

dancooke: Hashing is unlikely to be faster than bucket sort for integers values since bucket sort is essentially a hashing where the value is used as the hash. If the domain is large then the cost of allocating the array might outweigh the cost of computing the hash and managing allocations for the hash table. In any case, probably the difference will be small in practice and hashing is more generally applicable so I agree that it is probably the way to go, although I'd be curious to see some benchmarks against `qsort` to see the difference in practice.

As for why the mode is useful. For categorical values it is the only average the makes much sense (e.g. Boolean flag).

pd3: Sort algorithms have time complexity O(n log(n)) on average and some of them O(n^2) in the worst case whereas hash in O(n). Although sort can be faster for small in some situations, it does not scale up. 

More importantly though: Do you have a real usage scenario where this feature would be useful? Given that the range of a float tag can be about anything (e.g. between [0,1]) I am not convinced `mode` is that useful. It would have to be parametrized by the user, but is that worth supporting.

dancooke: [Counting sort](https://en.wikipedia.org/wiki/Counting_sort) - the variant of bucket sort that I am referring to - is worst-case O(n) time and space complexity.

I'm assuming that all numeric tag types are converted to `double` for the purpose of filter expressions? I agree that mode is not useful for real type input, but if the actual tag type was integral then the mode is useful, and as I mentioned is the only average that makes much sense for categorical types. For example, suppose I define a categorical tag where `0 = red`, `1 = blue`, `2 = green`, then the mean or median of a range of such values is meaningless, but the mode makes a lot of sense. I'm not sure what you mean by "It would have to be parametrized by the user".

pd3: I understand the concept. The question was aimed at did you add this to your pull request for the sake of completeness or do you have a real use case?

pd3:
This is now merged via c95468e. Thank you!

PR #1000, opened 2019-04-10
Fix address sanitizer complaints

daviesrob:
Fixes two complaints from address sanitizer:

* A memory leak that appeared in commit 0e79bdc

* A one-definition-rule violation in some of the plugins (notably setGT)

Not sure what's causing the second one, but making the offending symbols static fixes it.  They don't appear to be used outside the one `.c` file.

To reproduce the problems, configure bcftools with:

```
configure CFLAGS='-g -fsanitize=address' LDFLAGS='-fsanitize=address'
```

and then `make && make test`.

Note: Needs a C compiler that supports address sanitizer - either `clang` or a new enough `gcc`.

### DIFF
--- a/filter.c
+++ b/filter.c
@@ -53,8 +53,8 @@ static int filter_ninit = 0;
 #  define __FUNCTION__ __func__
 #endif
 
-uint64_t bcf_double_missing    = 0x7ff0000000000001;
-uint64_t bcf_double_vector_end = 0x7ff0000000000002;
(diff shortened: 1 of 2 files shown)


jmarshall:
The ODR violation is probably due to _filter.c_ being linked into both bcftools and a couple of the plugins.

pd3:
Thank you, now merged via d6cbf78c9ef07f2b9987a1d1598a0b2dca0ca061

PR #1002, opened 2019-04-12
Fix typo in warning message on enabling --gvcf option for mpileup

athos:


pd3:
Thank you for the fix.

PR #1034, opened 2019-05-31
Assorted bug fixes

daviesrob:
Fixes two warnings from `clang`.

Fix accidental read beyond the end of a string in vcfcall.c `set_samples()` when the line only contains the sample name.  If the program didn't crash and depending on what followed, this may have caused ploidy to be set incorrectly in some cases.

Add missing call to `hts_expand()` in mcall.c `mcall_constrain_alleles()` to fix an invalid write beyond the allocated region.  Also fix illegal pointer arithmetic.

Also adds an address sanitizer check to travis so problems like these can be detected more quickly.  It would help if the test harness didn't discard `stderr` quite so much though - you can tell that a test failed, but not why.  One solution to this might be to capture it into a file, and then print out the contents if the test happens to fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,11 @@ matrix:
     - os: osx
       compiler: clang
       env: USE_CONFIG=yes
+    # An optimised build with address, leak and undefined behavior checking
+    - os: linux
+      compiler: clang
+      sudo: required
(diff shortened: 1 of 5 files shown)


pd3:
Thanks for this!

PR #1057, opened 2019-07-31
Fix bcftools sort invalid input crash; fix compiler and make warnings

jmarshall:
Given an empty file _empty.vcf_ (or piped input from a program that failed, which is more likely), bcftools sort crashes due to not checking the return code from `bcf_hdr_read()`:

```
$ bcftools sort empty.vcf 
Writing to /tmp/bcftools-sort.QZKQj3
[E::bcf_hdr_read] Input is not detected as bcf or vcf format
Segmentation fault: 11
```

This fixes that to produce an error message and clean up instead.

Also suppress a number of `gcc-9` and/or `-Wextra` warnings. 

And avoid using `uname -o` in the Makefile, as that option does not exist on macOS or (some) BSDs. Instead check for `uname -s` values corresponding to -o "Msys", namely (according to [Wikipedia](https://en.wikipedia.org/wiki/Uname#Examples)) stuff matching /^(MINGW|MSYS).*$/. @valeriuo may wish to check that this still builds as expected on Windows platforms.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,8 +23,8 @@
 # DEALINGS IN THE SOFTWARE.
 
 CC       = gcc
-AR     = ar
-RANLIB = ranlib
+AR       = ar
+RANLIB   = ranlib
 CPPFLAGS =
 CFLAGS   = -g -Wall -Wc++-compat -O2
 LDFLAGS  =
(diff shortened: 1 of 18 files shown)


jmarshall:
(AppVeyor build failure is due to the use of `mkdtemp()`, already on **develop** from f1f261bdc5a44c22178a358f557ed90145a6ed49.)

valeriuo:
Looks good. No build warnings or errors on Win10 and all tests pass.

PR #1075, opened 2019-08-30
use /usr/bin/env perl

pvanheus:
Make vcfutils.pl script use the perl in the local environment rather than hard code /usr/bin/perl.


pd3:
Thank you.

PR #1083, opened 2019-09-13
Prepare for 64-bit positions

daviesrob:
This does not add 64-bit position support, it just ensures bcftools will build without warnings (using gcc/clang -Wall) when built against both 32-bit and 64-bit position branches of HTSlib.

Makes various printf()s use PRId64 for positions.  The values are cast to (int64_t) so both 32-bit and 64-bit versions work.

### DIFF
--- a/consensus.c
+++ b/consensus.c
@@ -333,7 +333,7 @@ static void unread_vcf_line(args_t *args, bcf1_t **rec_ptr)
 {
     bcf1_t *rec = *rec_ptr;
     if ( args->vcf_rbuf.n >= args->vcf_rbuf.m )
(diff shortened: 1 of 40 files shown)


pd3:
Thank you for this

daviesrob:
Thanks for merging, but unfortunately you were a bit too quick here, as two more fixes need to be added due to changes in bcftools since this was originally written.  Would you mind if I push in another commit to add them?

pd3:
Of course, sorry. I can revert if you want, but a separate commit is good too.

daviesrob:
No need to revert, the problem is now fixed via c17a908

pd3:
Thank you.

Issue #1068, opened 2019-08-14
Fixref questions


bgb-ipl:
Hi,

I have some questions about the fixref plugin :

1) Looking at the code, I can see that "TOP-compatible" and "BOT-compatible" outputs in the logs are not counts but can only take 0 or 1 value. Could you explain what do these flag mean ?

2) Using Illumina manifest, I can know TOP and BOT alleles but I don't see how I can provide such information to bcftools. For example, how do I represent a TOP [A/T] compared to a BOT [A/T] in the vcf file ?

Best,

pd3:
1. this is a check if all REF/ALT alleles in the VCF are compatible with either TOP or BOT notation: the values mean "true" (1) or "false" (0)

2. I don't understand the question. In VCF we use neither TOP nor BOT convention, we use the reference strand, defined by the reference genome stored in a fasta file. The TOP/BOT conventions enable to represent alleles locally, in the absence of a reference genome.

bgb-ipl:
1. Ok, so that's why we get 0 for both if we mix TOP and BOT in the same vcf file.

2. Sorry, the question was confusing. For example, if I know, following the Illumina manifest, that I have some BOT [T/A] alleles with 2 individuals T/T and A/A (but don't know on which strand). If I have to represent this variant in the vcf for fixref, I should represent it like :

```
#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	Ind1	Ind2
1	97547947	snp1	T	A	.	.	.	GT	0/0	1/1
```
OR
```
#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	Ind1	Ind2
1	97547947	snp1	A	T	.	.	.	GT	1/1	0/0
```

In the end, I have the same results for both representation :+1: 
```
#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	Ind1	Ind2
1	97547947	snp1_ref	T	A	.	.	.	GT	0/0	1/1
```
But the first representation tells BOT compatible => 1 and the second representation tells TOP compatible => 1. 

So, for statistics and checking purpose it seems better to represents : 
 - TOP [A/T] alleles as A=>REF and T=>ALT so that fixref will tell TOP compatible 
 - BOT [T/A] alleles as T=>REF and A=>ALT so that fixref will tell BOT compatible

maegsul:
Hi, I have the same problem described in the second question. 

[...]

PR #1086, opened 2019-09-20
Update makefile dependencies

jmarshall:
Reapply some makefile fixes that were overwritten during an earlier merge, and bring the dependencies right up to date with any recent `#include` changes in the source code.

Also a tiny bug fix: produce an error message instead of the segfault mentioned in https://github.com/samtools/bcftools/issues/1068#issuecomment-523099376 item number 1.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,9 +75,6 @@ MISC_SCRIPTS = \
     misc/run-roh.pl \
     misc/vcfutils.pl
 TEST_PROGRAMS = test/test-rbuf test/test-regidx
-PLUGINC =
-PLUGINS =
-PLUGINM =
 
 ALL_CPPFLAGS = -I. $(HTSLIB_CPPFLAGS) $(CPPFLAGS)
 ALL_LDFLAGS  = $(HTSLIB_LDFLAGS) $(LDFLAGS)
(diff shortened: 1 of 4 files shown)


jkbonfield:
It makes me wonder why we never added a "make depend" target to automate this.  It doesn't matter what tools it uses as it's not a dependency itself for people building the package, just for those maintaining it.

I did this with the Staden Package.  I had a depend target which ran the makedepend tool (X11, but various gcc options and dump out similar things) and then had a bunch of sed lines to remove system paths and turn absolute paths into relative ones before editing the Makefile itself.

jmarshall:
Because nobody advocating for doing this has actually done it :smile:

It's complicated by the `$(htslib_foo_h)` variables transmitting the dependencies across the projects (which IMHO is a great solution to htslib/samtools/bcftools coordinated separate repos vs. installed htslib building), but that's not insurmountable. There's previous discussions on all this and adding a `make depend` but I don't have time to look them up just now.

Tools-wise, I keep an eye on this using a horrifying script (usage: [./fmt-includes Makefile > Makefile2; diff -u Makefile{,2} | eyeballs](https://gist.github.com/jmarshall/d23a7dfae881a0f293fa0a70deaff430)) that could be contributed to the _maint_ repo one day, but I haven't got around to it.

pd3:
Thank you for this

PR #1098, opened 2019-10-30
header flag capitalization is switched in FAQ

gruberjd:
http://samtools.github.io/bcftools/howtos/FAQ.html#incorrect-nfields

The suggested command to query the header uses the -H (meaning --no-header), and the one to query the data uses -h (meaning --header-only).


pd3:
This pull request asks to merge 257 commits into master. Please make it against the develop branch.

PR #1099, opened 2019-11-01
remove cast warning for plugins/split-vep.c

lindenb:
This small PR removes warning ( realloc without cast) when compiling split-vep.c.

```
plugins/split-vep.c: In function ‘init_data’:
plugins/split-vep.c:248:21: warning: request for implicit conversion from ‘void *’ to ‘char **’ not permitted in C++ [-Wc++-compat]
         args->field = realloc(args->field,args->nfield*sizeof(*args->field));
                     ^
plugins/split-vep.c:302:30: warning: request for implicit conversion from ‘void *’ to ‘char *’ not permitted in C++ [-Wc++-compat]
             args->column_str = realloc(args->column_str, olen + nlen + 2);
                              ^
plugins/split-vep.c:348:20: warning: request for implicit conversion from ‘void *’ to ‘int *’ not permitted in C++ [-Wc++-compat]
             column = realloc(column,args->nannot*sizeof(*column));
                    ^
```

### DIFF
--- a/plugins/split-vep.c
+++ b/plugins/split-vep.c
@@ -245,7 +245,7 @@ static void init_data(args_t *args)
         char tmp = *ep;
         *ep = 0;
         args->nfield++;
-        args->field = realloc(args->field,args->nfield*sizeof(*args->field));
+        args->field = (char**)realloc(args->field,args->nfield*sizeof(*args->field));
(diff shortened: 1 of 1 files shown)


pd3:
Thank you

PR #1102, opened 2019-11-05
Update print statements to be both Python 2 and 3 compatible

jmarshall:
I haven't constructed any test data to run these on to see if there are other Python 3 issues in the less trivial parts of the code, but fixing the `print` statements hits the low-hanging fruit and makes all of

```
python2.7 misc/guess-ploidy.py
python2.7 misc/plot-roh.py
python3 misc/guess-ploidy.py
python3 misc/plot-roh.py
```

compile successfully and print out the usage as expected.


pd3:
Thank you for this.

PR #1103, opened 2019-11-06
Fix configure option thinko

jmarshall:
`@PLUGINPATH@` wasn't used anywhere, so `--with-bcf-plugin-path` didn't work.

Fix this by remembering to set make's `$(pluginpath)` from the `AC_SUBST` variable (and rename the variable to follow `@bcf_plugindir@`'s lead).

### DIFF
--- a/config.mk.in
+++ b/config.mk.in
@@ -50,7 +50,8 @@ PERL_LIBS = @PERL_LIBS@
 
 PLATFORM   = @PLATFORM@
 PLUGINS_ENABLED = @enable_bcftools_plugins@
-plugindir = @bcf_plugindir@
+plugindir  = @bcf_plugindir@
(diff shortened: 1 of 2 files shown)


jmarshall:
(Appveyor failed due to an unrelated _hts.dll_ linking error — `final link failed: File truncated` — that's presumably fallout from the `-fvisibility=hidden` stuff.)

daviesrob:
@jmarshall Looks like it was just a random failure when writing a file.  I told it to rebuild the PR and it's quite happy now.

jmarshall:
Thanks. I did wonder about that but I couldn't see a retry button…

daviesrob:
I fear that I may be the only person who can see the 'Rebuild PR' button.

pd3:
Thank you for this.

PR #1116, opened 2019-11-25
NEWS line-wrapping and typos / Fix some GCC warnings

jmarshall:
Rewrap the lines to 80 columns for legibility on narrower windows, and a couple of typo fixes.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,12 +1,14 @@
 ## Release a.b
 
-* Numerous bug fixes, usability improvements and sanity checks were added to prevent common user errors.
+* Numerous bug fixes, usability improvements and sanity checks were added
+  to prevent common user errors.
 
-* The -r, --regions (and -R, --regions-file) option should never create unsorted VCFs or duplicates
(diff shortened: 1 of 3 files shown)


jmarshall:
Also another commit to avoid “needs a cast” warnings:

> request for implicit conversion from ‘void *’ to ‘htsThreadPool * {aka struct <anonymous> *}’ not permitted in C++ [-Wc++-compat]

as seen in [non-configure builds](https://travis-ci.org/samtools/bcftools/jobs/616716179). `-Wc++-compat` is in the default Makefile's $CFLAGS, but if you run `configure` that is replaced by the usual `-Wall -g -O2` hence it was easy to not notice these warnings in local build trees.

jmarshall:
Forgot to mention that there may be a bug nearby the _vcfconcat.c_ one: in this `args->files == 0` case, `args->tpool.qsize` is left at 0 which perhaps leads to `hts_set_opt` setting not very useful multi-threading. I don't know what `qsize` is supposed to be set to… (and this may not be the time to alter this in any case).

pd3:
Thanks for this

PR #1119, opened 2019-11-27
Allow for absolute paths on Windows

valeriuo:
https://docs.microsoft.com/en-us/dotnet/standard/io/file-path-formats

### DIFF
--- a/vcfplugin.c
+++ b/vcfplugin.c
@@ -211,7 +211,12 @@ static void *dlopen_plugin(args_t *args, const char *fname)
 
     void *handle;
     char *tmp;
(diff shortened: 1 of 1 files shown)


pd3:
Thank you

jmarshall:
The Windows path stuff should only be done for `#ifdef _WIN32` etc.

Also the proposed code reads from uninitialised memory when `fname` is the empty string. It is always amazing what innocuous looking code can have potential security implications! :smile:

pd3:
Thanks for noticing. Addressed by 2e2cc71, plus now fixed a typo in that commit e4ef324.

PR #1132, opened 2020-01-02
Updated the list of OS packages.

jkbonfield:
See https://github.com/samtools/www.htslib.org/issues/24

I haven't updated the OSX ones as it was incomplete before.  I don't know which are standard OS packages and what package management tool was being recommended.  I assume MacOS people will just look at the other OSes and use whichever homebrew type equivalent system they wish, given there's not a standard package manager for the OS.


pd3:
Thank you for this

PR #1203, opened 2020-04-16
Update to +split-vep CSQ severity scale

McClatcheyM:
I've been working with the bcftools 1.10.2 and plugin +split-vep and using filter by consequence (+split-vep -s) on WGS data annotated with vep98
I am finding there are variants passing through the consequence filter in to my output files that should not be there i.e. are less severe than the stated minimum threshold.

I note that the VEP consequence terms (Calculated Consequences [https://m.ensembl.org/info/genome/variation/prediction/predicted_data.html]) 
have been updated since last review of the plugin severity scale 

I've attempted to amend this in split-vep.c with new terms have added and retired terms retained i.e. to account for users of previous vep versions. Any amendments/rejections/suggestions would be welcomed

*NB. This is my first time creating a fork/pull-request. Please forgive any errors in edit or ettiquette

### DIFF
--- a/plugins/split-vep.c
+++ b/plugins/split-vep.c
@@ -119,21 +119,27 @@ static const char *default_severity(void)
         "# Default consequence substrings ordered in ascending order by severity.\n"
         "# Consequences with the same severity can be put on the same line in arbitrary order.\n"
         "intergenic\n"
+	     "feature_truncation feature_elongation\n"
+	     "regulatory\n"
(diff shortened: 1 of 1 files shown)


pd3:
This is great, thank you for this. Now merged with minor adjustments via https://github.com/samtools/bcftools/commit/36536eb54c953c6d646e3c17ba386d0c654c3f42

PR #1205, opened 2020-04-21
Fix bcftools build on OpenBSD and Illumos

jmarshall:
Minor fixes needed to make bcftools build on these minor platforms. BCFtools companion to similar previous PRs samtools/htslib#1011, samtools/samtools#1151, and samtools/samtools#1165.


pd3:
Thank you

PR #1236, opened 2020-06-08
Fix convert bug and avoid other new Clang warnings

jmarshall:
Compiling bcftools with Clang **master** (the upcoming release 11) produces:

```
vcfnorm.c:1427:13: warning: misleading indentation; statement is not part of the previous 'if' [-Wmisleading-indentation]
            nret /= nsmpl;
            ^
vcfnorm.c:1426:13: note: previous statement is here
            if (nret<0) continue; /* format tag does not exist in this record, skip */ \
            ^
vcfnorm.c:1474:17: warning: misleading indentation; statement is not part of the previous 'if' [-Wmisleading-indentation]
                nret /= nsmpl;
                ^
vcfnorm.c:1473:17: note: previous statement is here
                if (nret<0) continue; /* format tag does not exist in this record, skip */ \
                ^


convert.c:788:26: warning: implicit conversion from 'int' to 'float' changes value from -2147483647 to -2147483648 [-Wimplicit-const-int-float-conversion]
            if ( ptr[j]==bcf_int32_vector_end ) break;
                       ~~^~~~~~~~~~~~~~~~~~~~
../htslib/htslib/vcf.h:1291:31: note: expanded from macro 'bcf_int32_vector_end'
#define bcf_int32_vector_end (-2147483647)  /* INT32_MIN + 1 */
                              ^~~~~~~~~~~


test/test-regidx.c:176:41: warning: implicit conversion from 'int' to 'float' changes value from 2147483647 to 2147483648 [-Wimplicit-const-int-float-conversion]
    *beg = min + (float)b * (max-min) / RAND_MAX;
                                      ~ ^~~~~~~~
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/stdlib.h:105:18: note: expanded from macro 'RAND_MAX'
#define RAND_MAX        0x7fffffff
                        ^~~~~~~~~~
test/test-regidx.c:177:43: warning: implicit conversion from 'int' to 'float' changes value from 2147483647 to 2147483648 [-Wimplicit-const-int-float-conversion]
    *end = *beg + (float)e * (max-*beg) / RAND_MAX;
                                        ~ ^~~~~~~~
```

The _vcfnorm.c_ warnings are trivially fixed.

The _convert.c_ warning indicates a bug, as `ptr[j]` is float data, not integer.

The `random()` stuff is similar to that addressed in samtools/htslib#1083, except that `random()` ranges up to 2<sup>31</sup>-1 not `RAND_MAX`. This is a bug fix on Windows, where `RAND_MAX` is much smaller.

### DIFF
--- a/convert.c
+++ b/convert.c
@@ -785,8 +785,8 @@ static void process_gp_to_prob3(convert_t *convert, bcf1_t *line, fmt_t *fmt, in
         int j;
         for (j=0; j<n; j++)
         {
-            if ( ptr[j]==bcf_int32_vector_end ) break;
-            if ( ptr[j]==bcf_int32_missing ) { ptr[j]=0; continue; }
+            if ( bcf_float_is_vector_end(ptr[j]) ) break;
+            if ( bcf_float_is_missing(ptr[j]) ) { ptr[j]=0; continue; }
             if ( ptr[j]<0 || ptr[j]>1 ) error("[%s:%"PRId64":%f] GP value outside range [0,1]; bcftools convert expects the VCF4.3+ spec for the GP field encoding genotype posterior probabilities", bcf_seqname(convert->header,line),(int64_t) line->pos+1,ptr[j]);
             sum+=ptr[j];
         }
--- a/test/test-regidx.c
+++ b/test/test-regidx.c
@@ -172,9 +172,9 @@ void test_custom_payload(void)
 
 void get_random_region(uint32_t min, uint32_t max, uint32_t *beg, uint32_t *end)
 {
-    long int b = random(), e = random();
-    *beg = min + (float)b * (max-min) / RAND_MAX;
-    *end = *beg + (float)e * (max-*beg) / RAND_MAX;
+    uint64_t b = random(), e = random();
+    *beg = min + (b * (max-min)) / INT32_MAX;
+    *end = *beg + (e * (max-*beg)) / INT32_MAX;
 }
 
 void test_random(int nregs, uint32_t min, uint32_t max)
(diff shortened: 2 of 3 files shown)


pd3:
The man page for random() at my system claims to return *successive pseudo-random numbers in the range from 0 to RAND_MAX* and claims to confirm to POSIX.1-2001, POSIX.1-2008, 4.3BSD.

jmarshall:
The man page on your system is incorrect, and has [recently been corrected](https://lore.kernel.org/linux-man/02185B3C-F6E5-4674-BB34-E549E04C5B37@glasgow.ac.uk/T/#m873245586a2677f870467f4950117bc01b6094ce).

jkbonfield:
This has surprised me as I also thought `RAND_MAX` was correct.  The relevant POSIX specs can be seen here:

https://pubs.opengroup.org/onlinepubs/9699919799/functions/random.html
https://pubs.opengroup.org/onlinepubs/9699919799/functions/rand.html

Summarising that, `rand()` is 0 to RAND_MAX and RAND_MAX is just guaranteed to be >= 32767.  It doesn't have to be 2^31-1.  `random()` is 0 to 2^31-1 and doesn't mention RAND_MAX anywhere.  Learn something new every day!

pd3:
It is not just the man page, but also the header files, they made sure RAND_MAX is defined there accordingly. 
```
/usr/include/stdlib.h:

/* The largest number rand will return (same as INT_MAX).  */
#define RAND_MAX    2147483647

/* Return a random long integer between 0 and RAND_MAX inclusive.  */
extern long int random (void) __THROW;
```

jmarshall:
Read the [man-pages mailing list thread](https://lore.kernel.org/linux-man/02185B3C-F6E5-4674-BB34-E549E04C5B37@glasgow.ac.uk/), or just the first message at that link.

They happen to be the same on Linux; I expect this is by design in glibc. But `RAND_MAX` is lower on other platforms. Notably on Windows, which is a platform on which bcftools is built.

That header file comment on `random()` is unchanged since 1995 or before, is misleading, and was missed when [glibc bug 7003](https://sourceware.org/bugzilla/show_bug.cgi?id=7003) led to the info documentation being corrected in 2013.

pd3:
Looks like a very slow process. Mine version of the documentation is 2017-09-15 and still refers to RAND_MAX

PR #1249, opened 2020-06-23
add scatter plugin

freeseek:
This scatter plugin should work as a convenient inverse to `bcftools concat` and provide a solid framework for parallelization in cloud pipelines together with `bcftools merge` and the split plugin.


pd3:
Merged, thank you for this!

PR #1254, opened 2020-07-03
improvements to the mendelian plugin

freeseek:
This patch (i) adds the ability for the mendelian plugin to read ped files, including ped files that might contain samples not present in the VCF, and (ii) adds the ability for the mendelian plugin to output VCFs with mendelian errors annotations


pd3:
Thank you for the pull request. May I propose a small change: 

```
-l, --list [+x]       list consistent (+) or inconsistent (x) sites [DEPRECATED, use -m instead]
-m, --mode [+xas]     output mode (the default is `-m s`):
                         + .. list consistent sites
                         x .. list inconsistent sites
                         a .. add INFO/MERR annotation with the number of inconsistent samples
                         s .. print a text summary with the number of errors per trio
```
where the `-l` option will continue to be functional, but a warning will be printed.

pd3:
I made a few more changes to allow listing of uninformative sites as well, in the process renamed `-m s` to `-m c`. Thank you for this.

PR #1295, opened 2020-09-01
Fix buffer (read) overrun in bcftools consensus.

jkbonfield:
For alleles with "END" the variant may be "imprecise" meaning that the
length of the REF allele differs to the END-POS+1 value.

We fill out consensus based on the length of REF and not rec->rlen.

An example file to trigger the issue is test/consensus.vcf.


pd3:
Thank you. Hopefully this is a better fix: dc26549.

PR #1337, opened 2020-11-03
Appveyor updates

daviesrob:
Upgrade the image used by Appveyor.  This is necessary to get the new keys and pacman version required for MSYS2.

Also remove a restriction that made it only build the develop branch, so that other branches can be checked before turning them into pull requests.


pd3:
Thank you

PR #1351, opened 2020-11-24
Miscellaneous minor fixes

jmarshall:
Motivated initially by `make test` producing an untracked file _test/norm.3.fa.fai_, several unrelated minor fixes:

* Fix some typos in error messages and file output (including corresponding fixes in expected output files);

* _test/*.fa.fai_ files are regenerated by the various test commands that use _test/*.fa_, so remove them from the repository, remove them in `make testclean`, and git-ignore them;

* Add `#include <strings.h>` to a couple of source files that are missing it. Other source files that use `strcase*()` functions \#include this header themselves, so do so for these two source files too.


pd3:
Thank you

Issue #1353, opened 2020-11-27
bcftools merge issues with multiple alternative alleles at one site (and with files missing that record)


dlaehnemann:
I have compiled a curious test case for `bcftools merge`:
[samples.tar.gz](https://github.com/samtools/bcftools/files/5607494/samples.tar.gz)

This contains `vcf.gz` files and respective indices for four samples. `sample1` contains only the header and no records. `sample2` and `sample4` contain the same alternative allele at the same site.`sample3` contains a different alternative allele at the same site. Here's what I get with `bcftools 1.11` when I try to merge the respective samples (it always prints the header and then fails with different messages, I skip the header here for brevity).

**Merging all samples**:
```
bcftools merge sample1.vcf.gz sample2.vcf.gz sample3.vcf.gz sample4.vcf.gz
Segmentation fault
```

**Merging all samples excluding the one with no record**:
```
bcftools merge sample2.vcf.gz sample3.vcf.gz sample4.vcf.gz
free(): invalid next size (fast)
Aborted
```
The same happens when excluding sample2 or sample4, so whenever we have the combination of different alternative alleles.

**Merging only those with the same alternative allele works**:
```
bcftools merge sample2.vcf.gz sample4.vcf.gz
```

Interestingly, this does not fail when done on full files with further records. Instead, it prints out a broken VCF record:
```
10      123256280       .       T       G,C     .       .       PROB_ABSENT=66.7183,21.6286;PROB_ARTIFACT=inf,inf;PROB_PRESENT=9.2461e-07,0.0299518     DP:AF:OBS:SB:ROB        .:.:.,.:.,.:.,. 3,.:1,.:3Vp:.,.:.,.     .,984:.,0.00101626:.,5:.,.:.,.  3,.:0.75,.:2Vp:.,.:.,.
```
I see two problems here:
1. The first sample has only one dot each for `DP` and `AF`, with no comma separator, whereas the other samples each have two entries.
2. The `OBS` entry in the third sample is truncated after the first digit, even though the header specifies it as `String`.

I'll try to compile a minimal working example to showcase this latter behaviour, as well, and post it here.

dlaehnemann:
Here comes a tar with BCF files that reproduce the last point mentioned above:
[samples.extra_record_bcfs.tar.gz](https://github.com/samtools/bcftools/files/5607778/samples.extra_record_bcfs.tar.gz)

The command for merging (with `bcftools 1.11`) is:
```
[...]

PR #1355, opened 2020-11-30
Enlarge merge_format_string() array for BCF_VL_A/BCF_VL_R too

jmarshall:
In `merge_format_string()`, the second loop does not enlarge `nmax` — leading to valgrind detecting invalid writes when the function later writes to an insufficiently large `tmp_arr`. See suggested fix; this appears to be the cause of #1353.

This PR also lengthens some strings in _test/merge.6.b.vcf_ so that the test case exercises the problem.


pd3:
Thank  you

Issue #1343, opened 2020-11-11
Best way to filter by F_MISSING in samples with a large number of alternate alleles?


CreRecombinase:
Hi, I'm QCing a vcf with a large number of samples, and I'm finding that when  trying to filter by F_MISSING I keep running in to `TODO: the GT fmt_type is not int8`, which I've tracked down (I think) [to this line in filter.c](https://github.com/samtools/bcftools/blob/78003de6dc5f24257a33230040061b464a442d2b/filter.c#L1063).  I have some multiallelic sites that have >60 alternate alleles, which I'm guessing is causing a promotion above `int8` for the GT field.  Do I need to split the multiallelic sites, do my filtering and rejoin them, or do I totally misunderstand the issue?  Also, is this the kind of TODO that's in the "good first contribution" category?

pd3:
Many ALT alleles are now supported with F_MISSING. Thanks for the issue.

PR #1356, opened 2020-11-30
 remove error before switch when fmt->type!=int8_t

CreRecombinase:
Looks like when #1343 was resolved this line wasn't removed, so non int8 `GT` never make it to the switch statement.


pd3:
Indeed, sorry about that.

PR #1363, opened 2020-12-18
DRAFT: BAQ avoidance

jkbonfield:
Not every call to BAQ is needed, indeed not every call is beneficial.

This is a work-in-progress DRAFT PR, but I'm making it as a reminder to my post-Christmas self. :-)

Basically it replaces the always-call-`sam_prob_realn` code with some analysis on indels and soft-clips in the pileup constructor. Then during the pileup phase, any time we have reads overlapping a certain number of indels / soft-clips we'll trigger the realignment there.

The result is we shave off about 1/3rd of the CPU time for mpileup, while simulatenously slightly improving on SNP calling accuracy.  See the figure below on a low coverage (15x) copy of SynDip chr1:

image

Indel calling is marginally harmed, but it's very minimal and it's still better than minus-BAQ.

With BAQ
```
InDel TP     31820 /   29451
InDel FP      5198 /     596
InDel FN     14216 /   16585
```

With partial BAQ
```
InDel TP     31719 /   29388
InDel FP      5165 /     588
InDel FN     14317 /   16648
```

That's all indels and filtered indels (filtering is a *must* for indels from bcftools).


jkbonfield:
The full depth SynDip chr1 (approx 45x).

Time to mpileup|call was 49m28s, vs 105m16s for the standard mpileup method (although I can't recall whether that benchmark was before my other mpileup speed gains).

image

jkbonfield:
@pd3 There's lots of experimentation gone on here, and it needs more evaluation still.  However I think it's in a state suitable for you to explore at least.

By default BAQ is enabled as before.  In addition to the `mpileup -B` option to disable it, there is now a new `-D` option to partially disable it based on my heuristics.  I've no idea how well they work on multi-sample data (but I also don't know an appropriate truth set there) and nor do I know how well it'll work on low allelic frequency data sets or complex events, such as cancer data.  However for the "normal" case it's looking sound, particularly with deep data (but be sure to check both deep and shallow as depth is one of the heuristics used).

jkbonfield:
An example from GIAB HG002 evaluation, including GATK 4 for fun.  (All tools with some complex filtering rules applied.)  The blue bars by the key are wall-clock time.  CPU time is marginally more for bcftools (mpileup|call), and about 80% more for GATK due to threading.

image

jkbonfield:
Latest batch of changes, focusing on Indel calling accuracy, but also demonstrating the effect of changing several default parameters.  I've left IMF at 0.05, but my "filtered" values in the tables below has it at 0.1.  I think it's maybe worth changing that default too, but it's worthy of debate as it's something we can trivially post-filter on.

This experiment is 60x HG002 GIAB sample.  Tests are still running, so this is just ~150Mb worth of chr1.  When it's finished I'll produce some plots.

Here FP = false positives, FN = false negatives and TP = true positives, with true/false defined as matching position and sequence.  GT is the genotype errors, so the call was correct but labelled 0/1 when it should have been 1/1 (or vice versa).  An incorrect GT but otherwise correct is still included in the TP column, so a more strict definition could be to subtract GT from TP and add it to FN/FP (or maybe half add/subtract if we deem a het vs hom call to be 1/2 correct, 1/2 incorrect).

```
# $DP = 2x depth = 60.
# $qual = as specified
snp_filt_expr="QUAL < $qual || DP>$DP"                                     
indel_filt_expr="IDV < 3 || IMF < 0.1 || DP>$DP || QUAL < $qual"            
```

1.11 defaults (-h100 -m1 -Q13)
```
SNP          Q>0 /   Q>=00 / Filtered   Q>=50 / Filtered
SNP   TP   98.89 /   98.89 /   98.88    98.62 /   98.62 
SNP   GT    0.12 /    0.12 /    0.12     0.10 /    0.10 
SNP   FP    0.91 /    0.91 /    0.89     0.36 /    0.35 
SNP   FN    1.11 /    1.11 /    1.12     1.38 /    1.38 
				                        
InDel TP   74.69 /   74.69 /   74.51    62.18 /   62.06 
InDel GT   12.42 /   12.42 /   12.37    10.93 /   10.90 
InDel FP   15.25 /   15.25 /    0.86    13.08 /    0.70 
InDel FN   25.31 /   25.31 /   25.49    37.82 /   37.94 
```

Not too bad on indels, but it's made a pretty poor job of indels, both in FN and FP.

1.11 -Q1
```
SNP          Q>0 /   Q>=00 / Filtered   Q>=50 / Filtered
SNP   TP   98.93 /   98.93 /   98.92    98.62 /   98.62 
SNP   GT    0.11 /    0.11 /    0.11     0.09 /    0.09 -
SNP   FP    0.96 /    0.96 /    0.94     0.35 /    0.33 -
SNP   FN    1.07 /    1.07 /    1.08     1.38 /    1.38 
				                        
InDel TP   75.76 /   75.76 /   75.57    62.55 /   62.42 
InDel GT   11.71 /   11.71 /   11.68    10.35 /   10.33 -
InDel FP   16.77 /   16.77 /    0.89    12.99 /    0.73 +
InDel FN   24.24 /   24.24 /   24.43    37.45 /   37.58 -
```

Dropping Q13 to Q1 is slightly beneficial. (Q0 isn't as it nullifies BAQ.)

1.11 -h500 -m2 -Q1 (Note -m2 is less strict than IDV<2; see unfilt col)
```
SNP          Q>0 /   Q>=00 / Filtered   Q>=50 / Filtered
SNP   TP   98.93 /   98.93 /   98.92    98.62 /   98.62 
SNP   GT    0.11 /    0.11 /    0.11     0.09 /    0.09 
SNP   FP    0.96 /    0.96 /    0.94     0.35 /    0.33 
SNP   FN    1.07 /    1.07 /    1.08     1.38 /    1.38 
				                        
InDel TP   95.86 /   95.86 /   95.74    94.94 /   94.87
InDel GT   19.29 /   19.29 /   19.27    19.18 /   19.16 ++
InDel FP    3.45 /    3.45 /    0.90     2.55 /    0.72 -
InDel FN    4.14 /    4.14 /    4.26     5.06 /    5.13 ----
```

Adding -h500 has a huge reduction to FN rates.  GT errors rise though.

Our filtering is IDV<3 so -m2 is irrelevant, but the unfiltered first column shows the remarkable drop in FP from ~17% to 3%.  The default of -m1 is just excessive.

This branch
```
SNP          Q>0 /   Q>=00 / Filtered   Q>=70 / Filtered
SNP   TP   99.15 /   99.15 /   99.15    98.75 /   98.75
SNP   GT    0.10 /    0.10 /    0.10     0.07 /    0.07 -
SNP   FP    1.02 /    1.02 /    1.00     0.32 /    0.30 -
SNP   FN    0.85 /    0.85 /    0.85     1.25 /    1.25 -
				                        
InDel TP   98.43 /   98.43 /   98.30    97.24 /   97.18
InDel GT    1.18 /    1.18 /    1.17     1.09 /    1.08 ----
InDel FP    2.22 /    2.22 /    1.11     1.61 /    0.91 +
InDel FN    1.57 /    1.57 /    1.70     2.76 /    2.82 --
```

Note using Q70 instead of Q50 filtering here to get ballpark same TP/FP SNP range.

Biggest change is the hew code has rescued those GT errors, dropping from 19% down to just 1%.  Indel FN has also more or less halved, and even unfiltered data is considerably more accurate.  The only downside is a slight increase in FP post-filtering (pre-filter it's better).

SNPs have marginally increased across the board vs 1.11 (including the modified options).  That's possibly an impact of better indel calling, but also maybe -D vs full-BAQ mode rescuing some missing calls.


This branch with improved filtering rules
```
SNP          Q>0 /   Q>=00 / Filtered   Q>=70 / Filtered
SNP   TP   99.15 /   99.15 /   99.13    98.75 /   98.73 
SNP   GT    0.10 /    0.10 /    0.10     0.07 /    0.10 +
SNP   FP    1.02 /    1.02 /    0.82     0.32 /    0.29 -
SNP   FN    0.85 /    0.85 /    0.87     1.25 /    1.27 +
				                        
InDel TP   98.43 /   98.43 /   98.41    97.24 /   97.69 
InDel GT    1.18 /    1.18 /    1.17     1.09 /    1.15 +
InDel FP    2.22 /    2.22 /    2.17     1.61 /    0.71 --
InDel FN    1.57 /    1.57 /    1.59     2.76 /    2.31 --
```

Finally with considerably more complex filtering rules we see a better tradeoff of FP vs FN is possible with indels, where the QUAL field is a poor proxy for accuracy (these filtering rules use it sparingly).

jkbonfield:
Latest round of tests, after upgrading to cope with PacBio CCS data (plots to follow).
Plots show QUAL filtering (from 0 to 200, although maybe clipped off the plot, in increments of 10), plus other basic filtering,
For SNPs that is `QUAL < $qual || DP>$dp*2` and for INDELs it is `IDV < 3 || IMF < 0.1 || DP>$dp*2 || QUAL < $qual`.  Far better filtering parameters are better, but I went with something far more likely to be in common usage here.

This is SynDip at 15x and 45x coverage.
Plots show bcftools 1.12 with the default parameters and with -h500 -m2 -Q1 (which are the new defaults for this PR).  Firstly that shows the impact of changing the defaults.

The thin lines with symbols, as shown in the keys, are false negative vs false positive using a very strict definition of correctness.  The variant must be the same coordinate, the same sequence, and the same genotype (ie 0/0, 1/1, 0/1, etc).  If the position and sequence is correct but the genotype incorrect, it adds half to FN and half to FP (as generally it's got one allele correct and one wrong).  The solid lines without symbols, but using the same colours, are without that penalty of requiring the correct genotype too.  It's noticably different on indels for the 1.12 as it had a high GT failure mode.

15x-snp
15x-indel

At 15x, SNPs don't show much difference between this PR and release 1.12 with updated default parameters.  With INDELs however there is a very significant shift once genotypes are checked, and even without that there is a shift towards fewer FN with the new code.

Note full vs partial BAQ has no impact on the indel caller.

45x-snp
45x-indel

At 45x, it's much the same as 15x only more exaggerated so it's easier to see the impact.  In particular partial BAQ with this PR is a significant reduction in FN rate for the SNP caller at the lower quality values (focusing on recall over precision).

jkbonfield:
This data set is GIAB HG002, subsampled to 15, 60 and 150x (from 300x).

15x-snp
15x-indel

At 15x there's very little in it with any tool / parameters for SNPs.  There is some marginal gain for the new code, but only visible if you zoom in a lot.  INDELs show the same gain in GT assignment accuracy we saw before, and once again is a little more sensitive even if we ignore the GT issue.

60x-snp
60x-indel

At 60x it's clear that partial BAQ is better than full BAQ for all SNPs except perhaps at very stringent filtering levels.  This deeper data really shows the impact of the new default parameters.  This is primarily -h500 as the default (-h100) misses around 30% of indels - presumably thinking they're polymerase stutter instead.

150x-snp
150x-indel

150x shows the same characteristics as 60x, only more so.  The reduction in false negatives for indels is particularly noticable.

jkbonfield:
Edit: I'll reanalyse the PacBIO CCS data set again now 1.12 has been released to do a proper side by side comparison once more, and post figures once I have them.  Meanwhile, the biggest impact in this PR is changing the defaults to be more appropriate for modern data and the bug fix to improve the indel genotype accuracy.  That was a relatively small change (using seqQ instead of indelQ, or vice versa -  I forget which way it went).

Finally, I recorded speeds too as this is what my original intention was.  Improving accuracy was just a happy fluke of analysing that my speedups weren't detrimental, and then researching the large source of missing indels.

GIAB HG002
--------------

| Depth  | Vers  |  Params         | CPU time (s) |
| ------ | ----- | --------------- | ------------ |
| 15x    | 1.12  |                 | 2060.46      |
| 15x    | 1.12  | -h500 -m2 -Q1   | 2044.61      |
| 15x    | mp    |                 | 2243.76      |
| 15x    | mp    | -D              | 1488.55      |
|  | | | |
| 60x    | 1.12  |                 | 5883.78      |
| 60x    | 1.12  | -h500 -m2 -Q1   | 5715.69      |
| 60x    | mp    |                 | 5956.72      |
| 60x    | mp    | -D              | 3327.77      |
| | | | |
| 150x   | 1.12  |                 | 14112.94     |
| 150x   | 1.12  | -h500 -m2 -Q1   | 12954.54     |
| 150x   | mp    |                 | 13071.56     |
| 150x   | mp    | -D              | 7887.00      |

CHM1/CHM13
----------

| Depth  | Vers  |   Params        | CPU time (s) |
| ------ | ----- | --------------- | ------------ |
| 15x    | 1.12  |                 | 2071.88      |
| 15x    | 1.12  | -h500 -m2 -Q1   | 2021.72      |
| 15x    | mp    |                 | 2206.46      |
| 15x    | mp    | -D              | 1496.16      |
|        |       |                 |              |
| 45x    | 1.12  |                 | 4994.60      |
| 45x    | 1.12  | -h500 -m2 -Q1   | 5029.81      |
| 45x    | mp    |                 | 5212.50      |
| 45x    | mp    | -D              | 2909.17      |

jkbonfield:
Now for some PacBio CCS results.  This data set was HG002 from GIAB again, and around 31x depth.  I've had to plot these in log scale (Y) because the range was so vast.  In particular bcftools 1.12 defaults really don't like CCS at all.

image

This is messy and hard to see due to the disparity in Y range and also there are multiple lines scrunched on top of one another.  Firstly, there are two clear parts - 1.12 with full BAQ (default) vs everything else.  1.12 defaults are just terrible for this data.  Fortunately 1.12 with -B to disable BAQ brings it into the mix with this PR.  It has slightly higher false positive rate, but the new defaults address that and it's then identical to this PR with defaults or -D (partial BAQ).  Why?  Because the new PR with full-baq or partial-BAQ is the same as no-BAQ due to a new maximum BAQ-length limit.

To force BAQ on, we need to use -M, set here to an arbitrary large value, and then we see a considerable improvement in FP and a slight improvement in FN.  As before I also show these figures with GT errors ignored, but the pattern is the same - just shifted.

image

Indels are a different picture, clustering in 3 locations.

Hard to see, but 1.12 with default params, both with and without BAQ, are in the top left corner.  The default parameters just don't work for indel calling on 1.12.

Next up is 1.12 with new defaults (-h500 -m2 -Q1), which we see rescues a major portion of the indels, but it's still missing > 10%.

This PR however is a very significant reduction in both FN and FP.  This is because I changed the BAQ parameters as the probabilities for the HMM state change are inappropriate for this technology.  They all line up together though as BAQ is forcibly on always for indels.  The without-GT-error case has just two lines visible, but it's mirroring the two groups above with data superimoposed, so the bottom left line is all the this-PR data.

Speed wise, there were a few surprises:

| Depth  | Vers  |   Params         | CPU time (s) | Avg Mem | Max Mem |
| ------ | ----- | ---------------- | ------------ | - | - |
| 30x    | 1.12  |                  |  81178.68    |  984.98 | 13888   |
| 30x    | 1.12  | -h500 -m2 -Q1    |  66833.25    | 1107.20 | 14710   |
| 30x    | 1.12  | -B               |  24111.57    |  425.10 |  5391   |
| 30x    | 1.12  | -B -h500 -m2 -Q1 |  12743.59    |  443.30 |  2992   |
| 30x    | mp    |                  |  13488.25    |   71.11 |    97   |
| 30x    | mp    | -M 99999         |  running
| 30x    | mp    | -D               |  13379.41    |   71.23 |    97   |
| 30x    | mp    | -D -M 99999      |   8380.38    |  525.77 |  3684   |

The 1.12 versions with BAQ were very slow and also huge significant memory up in computing BAQ, because the band size grows considerably making it a large slow matrix.  Without BAQ 1.12 is reasonably fast, although note the big difference between "-B" and "-B -m2 ...".  The -m2 there is removing a huge portion of candidate indels which get tested for and rejected.

This PR speed, with full and partial BAQ (without -M) are the same time, and they'd be much the same as no-BAQ too, due to the length limit being applied meaning no BAQ calcs were actually made.  They're slightly behind 1.12 -B, probably due to reserving memory for the potential BAQ qual values.

The suprise is `-D -M 99999` to forcibly enable some BAQ calcs to be permitted, where it thinks they'll help, bizarrely speeds things up!  I cannot explain why this is so.  It may warrant another test to check whether this is just an LSF abberation.  It's clear it did something as the memory increased and we also see it produce a different (better) line on the plots.  So maybe the default for -M should be higher.

I have some more tests running to see what full-BAQ `-M 99999` looks like.  I don't expect it to be the same as 1.12 for speed, as there is also a maximum size.  If BAQ would be too slow, it's now simply skipped.  I think it's set somewhere around 1/10th of a second for a BAQ calculation.  (On ONT testing of larger reads, I found some single BAQ calculations could take 30 mins, while others ran the system out of RAM - over 60GB!)

jkbonfield:
Closing as replaced by https://github.com/samtools/bcftools/pull/1474

PR #1378, opened 2021-01-20
Improve the Mann Whitney U test.

jkbonfield:
There are several main changes here.

1. A new implementation of the Mann Whitney U test.  The new code was
based on Wikipedia and reading of the GATK source (which very strongly
matches the Wikipedia article).  So far I am uncertain what the impact
of this has.

2. The U score is turned to a probability from 0 to 1 based on how
similar the distributions in a and b are.  However some tests are
asymmetric. Mapping Quality Bias (MQB) will have higher MQ for REF
than ALT as obviously sequencing differences reduce MQ.  In the code,
any absolute difference between U and mean reduces p towards 0,
however for MQB U substantially above mean is usually correct and U
substantially below is usually incorrect.  We have a function option
to avoid returning p values for U > mean.

3. An optional Z score alternative to U.  This returns the number of
standard deviations that U is away from mean, as either a +ve or
negative score.  This permits filters to explicitly test one or both
ends, eg -e "MQB < -4 || RPB < -5 || RPB > +5".  The numbers are also
an easier range to grasp than needing tiny numbers (eg "MQB < 1e-5 ||
RPB < 1e-6").

Currently the Z-score variant is #ifdefed out.  If enabled it will
generate new tags named MQBZ, BQBZ, RPBZ and MQSBZ.

### DIFF
--- a/bam2bcf.c
+++ b/bam2bcf.c
@@ -437,7 +437,7 @@ double calc_mwu_bias_cdf(int *a, int *b, int n)
     return pval>1 ? 1 : pval;
 }
 
-double calc_mwu_bias(int *a, int *b, int n)
+double calc_mwu_bias(int *a, int *b, int n, int left)
 {
     int na = 0, nb = 0, i;
     double U = 0, ties = 0;
@@ -461,6 +461,7 @@ double calc_mwu_bias(int *a, int *b, int n)
     if ( na==1 || nb==1 ) return 1.0;       // Flat probability, all U values are equally likely
 
     double mean = ((double)na*nb)*0.5;
+    if (left && U > mean) return 1; // for MQB which is asymmetrical
     if ( na==2 || nb==2 )
     {
         // Linear approximation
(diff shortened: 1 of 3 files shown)


jkbonfield:
Still a WIP, but here for early comment.  Obviously there are more experiments that need doing, and some unanswered questions such as how my algorithm differs to the original.  I'm a bit fuzzy on when to bail out with `HUGE_VAL` too.

The first 100Mb of Chr1 SynDip data set (full depth, so ~45x) produces this plot:

image

This is a bit confusingly drawn as log-log plot, but it gives better separation of FN vs FP as otherwise it's essentially a 1/x style curve with the values squished together a bit.

Each dot is an different $qual threshold.  Calls filtered via


```
bcftools view -e  "QUAL < $qual || DP>90 || MQB < 1e-5 || RPB < 1e-6"
```

The MQB and RPB are Mann Whitney U scores, which is what this PR rewrites.  So the difference between "Old U" and "New U" shows that impact.  The QUAL/DP only is without any MQB / RPB filtering, showing that filtering does help even with the old method.

Finally the GATK 4 calls are also shown (note  SNPs only here - it's way better than bcftools on indels, but neck and neck on SNPs for some QUAL levels).  The filtering options there are again either just QUAL/DP only, or the full one is:

```
QUAL < $qual || QD < 2 || FS > 40 || MQ < 20 || SOR > 6 || MQRankSum < -2 || ReadPosRankSum < -8 || INFO/DP>$DP
```

That was experimentally tweaked to get the best I could on this data set; as was done with bcftools values.

This is just step 1 of the overhaul.  I have some BAQ changes incoming too which pushes bcftools a bit further still towards bottom left corner.

jkbonfield:
A quick comparison of the old and new code shows some differences.

It's a reimplementation (in order to understand how it worked this was easiest).  So some differences that are equivalent, like counting number of equal matches `e` and less-than matches `l` rather than modifying `U` as it goes, but things that do change:

- I compute U from 1*(A<B count) + 0.5*(A==B count), while the old code is 1*(A>B count) + 0.5*(A==B count).   This is subtle as `nb` in the old code is the running total of b[i] going from 0 to n, so it's the count of b[j] where j<i.  The new code does the same but loops n to 0 instead so it's j>i.  The former is strictly what Wikipedia states, but I think this is irrelevant given the old code then folds any deviation from mean to 0..1 range in a symmetric fashion.

- New code does `if (na+nb <= 1)  return HUGE_VAL`.  The old code returned HUGE_VAL if either na or nb were 0, and 1.0 if either were 1.  We see this a lot in the output files with `MQSB=1` appearing on every non-variant line.  I'm not convinced this is a helpful number and we'd never filter to value 1.0.

- Old code having `na==1 || nb==1` in the above test means when both are 1 my `na+nb <= 1` doesn't trigger.  This is a genuine difference.  The old code returned 1.0 there; the new one would set mean = 0.5, variance = 0.25 and U would be 0, 0.5 or 1 depending on values, giving a result between 0.6 and 1.0.  A difference, but irrelevant for filtering I think.  Arguably we should have `na+nb <= X` for some higher threshold as we may want to not filter on such low sample sizes?

- My variance calculation uses the more accurate formula accounting for the number of ties.  I don't know how big an impact this makes, but it's cheap.  There is evidence the old code tried to do this, and indeed it accumulates the number of ties despite then ignoring it.  The commented out code appears to be wrong, which is why I assume it wasn't used.

- My code isn't using the `mann_whitney_1947` function that contains the fast lookup table.  This is because I'm permitting na and nb <= 1 and the lookup table starts from 2.  I could switch if I modified that check.

- I have an additional `if (var2 <= 0) return HUGE_VAL`.  This isn't needed in the old code as var2 can never be zero.  I *think* it's possible with the new code due to the tie correction.  It's harmless anyway as a zero variance means it's acceptable and shouldn't be filtered.  Arguably we should return 1.0 though (or 0.0 if Z score)?

I'm also unsure about the validity of the linear approximation for `na == 2 || nb == 2` case.  I copied it, but I don't know if I should have done.  I may try without it.  Especially if we ensure the lookup table version is available as then it'd be fast.  Looking at the maths I think it's quite a radically different result.  Some debugging info showing na/nb and the linear interpolation vs exact calc gives:

```
31/2    0.661290 0.730393
31/2    0.451613 0.420055
31/2    0.000000 0.061723
34/2    0.411765 0.384954
34/2    0.102941 0.098773
34/2    0.279412 0.236760
2/29    0.000000 0.049455
2/26    0.000000 0.053385
2/28    0.000000 0.048542
2/28    0.000000 0.048323
2/30    0.000000 0.047279
57/2    0.385965 0.341136
57/2    0.052632 0.074550
48/2    0.364583 0.319377
38/2    0.157895 0.139006
38/2    0.105263 0.001181
38/2    0.052632 0.069599
2/2     0.500000 0.417771
2/2     0.000000 0.511663
2/2     0.250000 0.511663
2/2     0.500000 0.417771
30/2    0.600000 0.645969
30/2    0.000000 0.054960
35/2    0.857143 0.945113
35/2    0.085714 0.094246

```

The 0.0000 cases there are worrying as they break filtering calculations.

jkbonfield:
Some plots showing log10(val) (X) vs probability of being false from false/(false+true) via the truth set.

The MQB+ is develop branch and MQB+U2 is this PR with the linear interpolation bit removed.

image

It's a bit hit and miss difference between the two.

Plus RPB:

image

This shows the impact of removing those 0 values, many of which are true (although the sample sizes aren't huge).  I'm trying again without the asymmetric hack again to see what impact it has with this fix in.  I'd prefer not to have it there as it's a cleaner description of the algorithm without it.

jkbonfield:
I've tried this with HG002 chr22 from GIAB at 15x, 30x and 60x depth.

My filtering is a bit complex, but it's trying to get the most from the data (calibrated off the SynDip data so not optimised on GIAB).

SNP exclusion, Z score: `QUAL < $qual || DP>$DP || MQBZ < -(3.5+100/(50+$qual)) || RPBZ > 2.5+100/(50+$qual) || RPBZ < -(2.5+100/(50+$qual))`
InDel exclusion, Z score: `IDV < 3 || IMF < 0.08 || DP>$DP || QUAL < $qual/5 || ((5+$qual/6)/(1.01-IMF) < $qual && IDV<(3+$DP*$qual/1000))`

SNP exclusion, U score: `QUAL < $qual || DP>$DP || MQB < 1e-4 || RPB < 1e-3`
InDel exclusion, U score: `IDV < 3 || IMF < 0.08 || DP>$DP || QUAL < $qual/5 || ((5+$qual/6)/(1.01-IMF) < $qual && IDV<(3+$DP*$qual/1000))`

image
15x: minimal difference

image
30x: new U/Z calcs are a bit ahead of the old code.

image
60x: considerable improvement.  (Although it'd be interestng to see BAQ-off with this PR)

The conclusion here is that this PR helps the existing U score considerably at high depth, while not being detrimental at low depth.  Z score is hard to compare as the filter rules differ (just what Z-score is equiv to `MQB < 1e-4`?) but it by and large follows a similar pattern.  The only surprise here is that at deep data the best discrimination of true vs false is to not use BAQ! 

[I have a variant (to come) which applied BAQ less frequently, only running it on data which it deems necessary for accurate calling.  It's not perfect, but it recovers the drops seen in 30x and 60x where no BAQ (-B) is better, while still working OK on the parts of the plot where BAQ helps.]

PR #1395, opened 2021-02-03
Reformat NEWS to be legible on an 80-column terminal

jmarshall:
Clarify `-O`/`--output-type` entry.
Remove mention of `--group-samples TAG:file` as that was not present in the previous 1.11 release.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -2,10 +2,11 @@
 
 Changes affecting the whole of bcftools, or multiple commands:
 
-* The output file type (-O, --output-type) needs not to be specified anymore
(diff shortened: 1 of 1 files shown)


jmarshall:
Thanks Petr.

BTW there's a warning on line 575 of _plugins/trio-dnm2.c_ (the new `denovo_allele` stuff): should it be `UINT8_MAX` or is this what the “making it fail deliberately” comment is about?

pd3:
Uh, not like this, this was not intended, my compiler did not warn me. Fixing now, thank you.

PR #1407, opened 2021-02-09
Continuous integration test updates

daviesrob:
Complete migration to cirrus-ci by disabling travis.

Rename .travis directory to vendor-neutral .ci_helpers

Change badge in README.md to a cirrus-ci one, and also add badges for Appveyor and the Github downloads count.


.gitattributes
```diff
@@ -5,6 +5,8 @@
 *.bam           -text diff=bam
 
 # Omit these files from release tarballs.
+/.appveyor.yml  export-ignore
+/.cirrus.yml    export-ignore
 .git*           export-ignore
-.travis*        export-ignore
+.ci_helpers*    export-ignore
```
jmarshall, line 11: At the risk of being very boring: the `*` here was so that it would match both _.travis_ and _.travis.yml_, so is no longer necessary; and `.travis` still appears in a comment in _.ci_helpers/clone_.

daviesrob: Good spot.  I have made some adjustments.

pd3:
Thank you

PR #1468, opened 2021-04-16
[index] Allow the index statistics to run just with the index file

### LINKED ISSUE #1418: Feature request: Allow reading statistics from index file without loading VCF file
This might seem like an atypical request, but when writing pipelines running in the cloud, sometimes I have the need to know in advance the number of samples and the number of markers inside a VCF file to make sure the proper amount of memory is allocated when requesting a VM.

If I don't have the number of markers at hand, I need to request a separate VM just to estimate this. This information is conveniently stored in the index file of a VCF but the `bcftools index --nrecords` and `bcftools index --stats` framework needs to open the VCF to work and sending the VCF could add a lot of extra cost if the VCF file is large. This has led me to the following little hack. Suppose I have an index file `in.bcf.csi`, then:
```
echo -e "##fileformat=VCFv4.2\n#CHROM\tPOS\tID\tREF\tALT\tQUAL\tFILTER\tINFO" | \
  bcftools view -Ob -o in.bcf
bcftools index --nrecords in.bcf
```
This allows to retrieve the number of variants from a binary VCF index without having to retrieve the binary VCF.

The only reason `bcftools index --nrecords` and `bcftools index --stats` need to open the VCF is to retrieve whether the VCF is a binary VCF or not and, if no variant counts are present in the index, whether this is due to the VCF file being empty or the index being an old index with no marker statistics.

It would be nice if there was a way to override this check to allow the extraction of the statistics from the index without needing the VCF file. Unrelated, it would be also nice if the index contained information about the number of samples in the VCF, so that there would be an alternative way to access this information without needing the VCF file.

valeriuo:
Some index statistics, such as the total number of records (`--nrecords`) and the number of records per chromosome (`--stats`) can be retrieved directly from the index file associated with a VCF/BCF file, without the need for the actual variant call file. This PR implements this behaviour, by making the VCF/BCF file optional.

There is still the aspect of HTSlib's `hts_open()` throwing an error, when the VCF/BCF is not found, but the error message can be discarded by redirecting `stderr` to `/dev/null`, thus obtaining a clean output.

Fixes #1418

### DIFF
--- a/vcfindex.c
+++ b/vcfindex.c
@@ -63,27 +63,37 @@ int vcf_index_stats(char *fname, int stats)
     const char **seq;
     int i, nseq;
     tbx_t *tbx = NULL;
+    bcf_hdr_t *hdr = NULL;
     hts_idx_t *idx = NULL;
 
     htsFile *fp = hts_open(fname,"r");
-    if ( !fp ) { fprintf(stderr,"Could not read %s\n", fname); return 1; }
-    bcf_hdr_t *hdr = bcf_hdr_read(fp);
-    if ( !hdr ) { fprintf(stderr,"Could not read the header: %s\n", fname); return 1; }
-
-    if ( hts_get_format(fp)->format==vcf )
-    {
-        tbx = tbx_index_load(fname);
-        if ( !tbx ) { fprintf(stderr,"Could not load index for VCF: %s\n", fname); return 1; }
+    if ( !fp ) {
+        tbx = tbx_index_load3(fname, NULL, HTS_IDX_SILENT_FAIL);
+	if ( !tbx )
+            idx = hts_idx_load3(fname, NULL, HTS_FMT_CSI, HTS_IDX_SILENT_FAIL);
+        if ( !tbx && !idx) {
+            fprintf(stderr,"Could not load file '%s', nor locate any index file associated with it.\n", fname);
+            return 1;
+        }
     }
-    else if ( hts_get_format(fp)->format==bcf )
-    {
-        idx = bcf_index_load(fname);
-        if ( !idx ) { fprintf(stderr,"Could not load index for BCF file: %s\n", fname); return 1; }
-    }
-    else
-    {
-        fprintf(stderr,"Could not detect the file type as VCF or BCF: %s\n", fname);
-        return 1;
+    else {
+        hdr = bcf_hdr_read(fp);
+        if ( !hdr ) { fprintf(stderr,"Could not read the header: %s\n", fname); return 1; }
+        if ( hts_get_format(fp)->format==vcf )
+        {
+            tbx = tbx_index_load(fname);
+            if ( !tbx ) { fprintf(stderr,"Could not load index for VCF: %s\n", fname); return 1; }
+        }
+        else if ( hts_get_format(fp)->format==bcf )
+        {
+            idx = bcf_index_load(fname);
+            if ( !idx ) { fprintf(stderr,"Could not load index for BCF file: %s\n", fname); return 1; }
+        }
+        else
+        {
+            fprintf(stderr,"Could not detect the file type as VCF or BCF: %s\n", fname);
+            return 1;
+        }
     }
 
     seq = tbx ? tbx_seqnames(tbx, &nseq) : bcf_index_seqnames(idx, hdr, &nseq);
@@ -94,11 +104,11 @@ int vcf_index_stats(char *fname, int stats)
         hts_idx_get_stat(tbx ? tbx->idx : idx, i, &records, &v);
         sum+=records;
         if (stats&2 || !records) continue;
-        bcf_hrec_t *hrec = bcf_hdr_get_hrec(hdr, BCF_HL_CTG, "ID", seq[i], NULL);
+        bcf_hrec_t *hrec = hdr ? bcf_hdr_get_hrec(hdr, BCF_HL_CTG, "ID", seq[i], NULL) : NULL;
         int hkey = hrec ? bcf_hrec_find_key(hrec, "length") : -1;
         printf("%s\t%s\t%" PRIu64 "\n", seq[i], hkey<0?".":hrec->vals[hkey], records);
     }
-    if (!sum)
+    if (!sum && fp)
     {
         // No counts found.
         // Is this because index version has no stored count data, or no records?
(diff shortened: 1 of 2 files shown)


vcfindex.c
```diff
@@ -63,27 +63,37 @@ int vcf_index_stats(char *fname, int stats)
...
-    if ( hts_get_format(fp)->format==vcf )
-    {
-        tbx = tbx_index_load(fname);
-        if ( !tbx ) { fprintf(stderr,"Could not load index for VCF: %s\n", fname); return 1; }
-    }
-    else if ( hts_get_format(fp)->format==bcf )
-    {
-        idx = bcf_index_load(fname);
-        if ( !idx ) { fprintf(stderr,"Could not load index for BCF file: %s\n", fname); return 1; }
+    if ( !fp ) {
```
jkbonfield, line 70: The error message from `hts_open` feels messy. 

It looks like `hts_idx_load3` works irrespective of whether we give it a bcf, bcf.csi, vcf.gz, vcf.gz.csi or vcf.gz.tbi.  I didn't check .bai as it looks like it's no longer capable of being written by bcftools so testing it is rather challenging.

I think the old code is confusing you too.  It uses .tbi for vcf and .bai/.csi for bcf, but vcf can have .bai/.csi too and it works just fine.  Did the old code simply not support .csi indices for vcf?  Furthermore it looks like `idx_index_load3` can be given a .tbi file anyway, so is `tbx_index_load3` a legacy interface now?  Furthermore it doesn't winge if given a .bcf or .vcf.gz file, unlike `tbx_index_load3`.

I'm not sure I understand why we try both APIs.  Just historical reasons, or am I missing something?

jmarshall: The error message from `hts_open` **is** messy: of all the API functions that can fail, opening files is the one that practically every calling program does actually check and print an error message for itself. IMHO `hts_open()` and the functions under it should not be printing messages themselves, but leave that to the caller — which knows the context.

Or even multiple spurious messy messages, e.g.,
```
$ samtools view htslib
[E::hts_hopen] Failed to open file htslib
[E::hts_open_format] Failed to open file "htslib" : Is a directory
samtools view: failed to open "htslib" for reading: Is a directory
```

jkbonfield: Ok what I was missing is the `--stats` output has chr length.  For some reason I had it in my head that this was in the undocumented CSI meta-data, but I can't find it with hd so I guess not.

However for `bcftools --nrecords` I can get the correct result for bcf, bcf.csi, vcf.gz, vcf.gz.csi and vcf.gz.tbi with the `htsFile *fp = hts_open(fname,"r");` line replaced by `htsFile *fp = NULL` (ie forced failure).  Furthermore commenting out the `tbx = tbx_index_load3...` line has no effect and it works just fine, so that's a red herring now it looks.

Sadly however in order get chromosome lengths from `--stats` we do have to open the main file up, which means we need to know whether to open it.  As there's no silent option (maybe something we should consider adding to htslib) nor is there a way of querying what type of index we loaded, so I think the only viable alternative for error-report free code therefore is checking file extensions.

Eg:

```
    char *d = strrchr(fn, '.');
    htsFile *fp = (d && (strcmp(d, ".tbi") == 0 ||
			 strcmp(d, ".bai") == 0 ||
			 strcmp(d, ".csi") == 0))
	? NULL
	: hts_open(fn, "r");
```

Assuming my code is correct, it'll not attempt to `hts_open` a file ending in .tbi, .bai or .csi, which avoids the bogus error message.

valeriuo: > I'm not sure I understand why we try both APIs. Just historical reasons, or am I missing something?

I followed the logic of the existing code, but I will try to do away with `tbx_index_load`.

> IMHO hts_open() and the functions under it should not be printing messages themselves, but leave that to the caller — which knows the context.

This is an option worth considering. My secondary solution was to add a silent mode (`s`) to `hts_open`, which would make it not log errors.

jkbonfield: Argh it's even more confusing.  I'd forgotten what a total nightmare our "twisty maze of index formats all like" really is.

So `tbx_index_load3` can read .tbi and .vcf.gz.csi but not .bcf.csi.  From vague recollection that's because CSI is actually two different formats - one for VCF and one for BCF.  (The bit that differs is, of course, the undocumented aux part.)

With `hts_index_load3` instead, it works on some .tbi files but not on others.  I have no clue what the difference is.   Eg (with hts_index_load3 commented out and tbx_index_load3 being used always):

```
$ bcftools index --stats /nfs/users/nfs_j/jkb/scratch/data/ALL.chr20.phase3_shapeit2_mvncall_integrated_v5a.20130502.genotypes.vcf.gz.tbi
[E::hts_hopen] Failed to open file /nfs/users/nfs_j/jkb/scratch/data/ALL.chr20.phase3_shapeit2_mvncall_integrated_v5a.20130502.genotypes.vcf.gz.tbi
[E::hts_open_format] Failed to open file "/nfs/users/nfs_j/jkb/scratch/data/ALL.chr20.phase3_shapeit2_mvncall_integrated_v5a.20130502.genotypes.vcf.gz.tbi" : Exec format error
20	.	1812841
```

With the vice-versa strategy, of tbx_index_load3 used and hts_index_load3 commented out we instead get:

```
$ bcftools index --stats /nfs/users/nfs_j/jkb/scratch/data/ALL.chr20.phase3_shapeit2_mvncall_integrated_v5a.20130502.genotypes.vcf.gz.csi
[E::hts_hopen] Failed to open file /nfs/users/nfs_j/jkb/scratch/data/ALL.chr20.phase3_shapeit2_mvncall_integrated_v5a.20130502.genotypes.vcf.gz.csi
[E::hts_open_format] Failed to open file "/nfs/users/nfs_j/jkb/scratch/data/ALL.chr20.phase3_shapeit2_mvncall_integrated_v5a.20130502.genotypes.vcf.gz.csi" : Exec format error
(null)	.	1812841
```

So it's lost the chromosome name.  Infact this happens to the .bcf file too.

Hence, sadly, I think we're going to have to check filenames and call the appropriate function if we don't want bogus error messages (eg as tbx_index_load3 gives when given the .bcf.csi file).

I'm starting to think the premise of this PR is wrong.  It's OK for `--nrecords`, but there's simply not enough data in the indices to work for `--stats`.

jkbonfield: Digging deeper, it is indeed CSI's schizophrenic nature which is causing that `(null)` for chromosome name.

vcf.gz.csi has the aux array as a tbi header block, which contains a mapping of IDs to names.  bcf.csi has a different format aux array that doesn't have this capability.

Practically speaking, the entire rest of the world uses .vcf.gz.tbi it seems as I believe we're the only package to have CSI indices?  So maybe the original requester wasn't aware of this.

valeriuo:
Working on a new solution.

Issue #1459, opened 2021-04-07
bcftools call ignores deletion with high coverage


LaraFuhrmann:
When executing bcftools call on the output of bcftools mpileup it sometimes fails to retain deletions with approriate coverage. 

For example, consider the following [bcf-file](https://github.com/samtools/bcftools/files/6270684/important_positions.bcf.gz), in particular the entry:
```
NC_045512.2     4179    .       AGG     AG      0       .       INDEL;IDV=752;IMF=0.516838;DP=1455;AD=1050,405;I16=360,690,405,0,15744,597840,14027,487333,63000,3.78e+06,24300,1.458e+06,12476,243234,10125,253125;QS=0.548652,0.451348;VDB=0.514509;SGB=-0.693147;MQSB=1;MQ0F=0  PL      0,52,61
```
When executing the command `bcftools call -Oz -mv --keep-alts --output indel_problem.bcf important_positions.bcf`, the deletion is lost and not listed in the output bcf-file, despite having 752 (out of 1455) raw reads supporting the deletion. 

Can we somehow keep the deletion?

jkbonfield:
There is a maximum depth parameter.  Places that are too deep get skipped.  This can be increased with the `-L` or `--max-idepth` option, which defaults to 250.

LaraFuhrmann:
From what I have seen (and also tested) bcftools call does not have the option `--max-idepth`. 

We used that option though when executing bcftools mpileup:
`bcftools mpileup -Ou -f "reference.fasta" --max-depth 1000000 --max-idepth 1000000 --annotate INFO/AD important_positions.bam --output important_positions.bcf`.

jkbonfield:
It's a `bcftools mpileup` option.

jkbonfield:
Sorry, I see what you mean.

It's hard to diagnose this without seeing the BAM file too.  The record in question has `PL      0,52,61` which, if my interpretation is correct, is claiming it's a REF match.  I don't know why mpileup concluded this, but I have some suspicions.

Please try `mpileup -h500`.  I found many quite deep indels are rejected because bcftools concludes they're far more likely to be sequencing errors, presumably caused to due polymerase slippage in homopolymers, which this is a small one.  The defaults for mpileup were written many years ago when sequencing artifacts were considerably more common than they are now, and for indels are not appropriate for modern data sets.  (I have a branch changing this and numerous other things to do with calling, but it's still a work in progress.)

LaraFuhrmann:
[...]

PR #1469, opened 2021-04-21
Add a --seed option to mpileup.

jkbonfield:
Random numbers are used by `ks_shuffle`, which is called in htslib's errmod code for sub-sampling the data in regions of > 255 depth.

This corrects the OS specific randomness seen in #1459.  It doesn't however solve the core of the problem (as that's in htslib).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -274,7 +274,7 @@ dist.o: dist.c dist.h
 cols.o: cols.c cols.h
 regidx.o: regidx.c $(htslib_hts_h) $(htslib_kstring_h) $(htslib_kseq_h) $(htslib_khash_str2int_h) regidx.h
 consensus.o: consensus.c $(htslib_vcf_h) $(htslib_kstring_h) $(htslib_synced_bcf_reader_h) $(htslib_kseq_h) $(htslib_bgzf_h) regidx.h $(bcftools_h) rbuf.h $(filter_h)
-mpileup.o: mpileup.c $(htslib_sam_h) $(htslib_faidx_h) $(htslib_kstring_h) $(htslib_khash_str2int_h) regidx.h $(bcftools_h) $(bam2bcf_h) $(bam_sample_h) $(gvcf_h)
+mpileup.o: mpileup.c $(htslib_sam_h) $(htslib_faidx_h) $(htslib_kstring_h) $(htslib_khash_str2int_h) $(htslib_hts_os_h) regidx.h $(bcftools_h) $(bam2bcf_h) $(bam_sample_h) $(gvcf_h)
 bam2bcf.o: bam2bcf.c $(htslib_hts_h) $(htslib_sam_h) $(htslib_kstring_h) $(htslib_kfunc_h) $(bam2bcf_h) mw.h
 bam2bcf_indel.o: bam2bcf_indel.c $(htslib_hts_h) $(htslib_sam_h) $(htslib_khash_str2int_h) $(bam2bcf_h) $(htslib_ksort_h) str_finder.h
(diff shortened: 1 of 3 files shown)


jmarshall:
LGTM… Also needs `$(htslib_hts_os_h)` added to the _mpileup.o_ dependencies in the _Makefile_, and ideally `‑‑seed` documented in the man page too.

jkbonfield:
Good points.  Thanks.

jkbonfield:
@pd3: Do we have any formal documentation on how to build our documentation?  I've regenerated .1 and .html (in a separate commit) from the .txt which is the source material, but it majorly differs.  I've no idea what local system was used before.

Locally at Sanger, I tried a2x on seq3 which works (it's used for this PR), but differs in HTML output.  I also tried seq4 and farm5, both of which fail:

```
cd doc && a2x -adate="2021-04-21 14:31 BST" -aversion=1.12-20-g54cb273+ --doctype manpage --format manpage bcftools.txt
a2x: ERROR: "xmllint" --nonet --noout --valid "/nfs/users/nfs_j/jkb/work/samtools_master/bcftools/doc/bcftools.xml" returned non-zero exit status 4
Makefile:323: recipe for target 'doc/bcftools.1' failed
make: *** [doc/bcftools.1] Error 1
```

The error in xmllint you see if you run it by hand:

```
I/O error : Attempt to load network entity http://www.oasis-open.org/docbook/xml/4.5/docbookx.dtd
/nfs/users/nfs_j/jkb/work/samtools_master/bcftools/doc/bcftools.xml:2: warning: failed to load external entity "http://www.oasis-open.org/docbook/xml/4.5/docbookx.dtd"
D DocBook XML V4.5//EN" "http://www.oasis-open.org/docbook/xml/4.5/docbookx.dtd"
                                                                               ^
/nfs/users/nfs_j/jkb/work/samtools_master/bcftools/doc/bcftools.xml:5: validity error : Validation failed: no DTD found !
<refentry lang="en">
```

If I remove the `--nonet` option, then it can validate the DTD and work.  However I don't know how to get a2x command line to run xmllint with that option removed.  We could however use `a2x -L` and editing the Makefile to use that works fine on the more modern systems (seq4 / farm5).

Should we be using `-L` always?  What does xmllint buy us?  It sounds rather like an internal diagnostic of Asciidoc, given we're not actually writing XML ourselves.  If it fails xmllint then surely asciidoc itself is broken?

Also does it matter if we have spurious differences caused by version changes in the auto-generated files, especially if we put them into their own commit so the actual source changes are obvious?

pd3:
This has been always done on vr-farm, but that was closed just a few days ago. I will look into this.

jkbonfield:
I rebuilt it using asciidoctor, as per the latest develop HEAD, and hardly surprisingly there are a lot of difference, but, IMO they're for the better.

Eg the a2x man page had:

```
       -S, --samples-file FILE
           File of sample names to include or exclude if prefixed with
           "&#94;". One sample per line. See also the note above for the -s,
           --samples option. The sample order is updated to reflect that given
```

The asciidoctor version has:

```
       -S, --samples-file FILE
           File of sample names to include or exclude if prefixed with "^".
           One sample per line. See also the note above for the -s, --samples
           option. The sample order is updated to reflect that given in the
```

I'll leave it up to you Petr if you wish to merge this PR as-is or just pick the first commit and do the man page rebuilding at release time.

pd3:
Would it be possible to remove the generated `.1` and `.html` files from the pull request?

jkbonfield:
@pd3 - rebased and removed the manual page changes (for generated output).

PR #1474, opened 2021-04-22
Mpileup revamp

### LINKED ISSUE #1446: bcftools call is detecting lots of REF reads that's not in the bam file (small test data set included)
When using bcftools to call variants on a virus genome, position `28253` has `DP4=6,86,162,25` and `AD` of `92,187`, but the bam file in the browser shows very few (if any) REF reads. I tried fiddling with `--no-BAQ`, `--min-MQ` and `bcftools call -m -P` per previous discussions but didn't see much difference. 

Command: 
`bcftools mpileup -Q 20 -d 100000 -L 100000 -a AD,DP,SP --min-ireads 20 --gap-frac 0.5 --no-BAQ -f MN908947.3.fa small_test.bam | bcftools call --ploidy 1 -c - | bcftools view -c 1 -Ov > small_test_no_BAQ.vcf`

[Test data](https://github.com/danrlu/debug_data/tree/main/bcftools_indel) to reproduce results from the command above.

Version: bcftools 1.12 from Conda

Possib related issues: #1276 #1348 

Any suggestions would be appreciated, and thank you for such a useful tool!!!

jkbonfield:
This is a rebased, reordered / squashed version of https://github.com/samtools/bcftools/pull/1363.

A summary of the changes are:

- BAQ is now tri-state; fully on (still the default), off, and partial.  Partial BAQ only runs BAQ on reads which overlap potentially problematic indels.  This is both a significant speed increase, but also an improvement in calling rates as BAQ can sometimes be detrimental so we only want to use it when we have a significant chance of fixing something.

- Added a new calc_mwu_biasZ function (Z-score) to replace calc_mwu_bias.  These are currently used if `MWU_ZSCORE` is defined in bam2bcf.h (which it is).  If we decide we prefer these metrics, then we can remove this ifdef and the old code, but for now there is a compilation choice to be made. (And no user way to configure.)  Alternatively it could become a run-time choice with a bit more work.  Please advise.

- Added MQBZ and RPBZ INFO fields to indels, for filtering on.

- Added a soft-clip bias SCBZ for SNPs, based on length of soft-clips near the SNP site, indicating potential misalignments. (Although questionable how helpful this is.)

- Very significant improvement to GT indel assignment accuracy.  In some data sets it reduces incorrect assignment by an order of magnitude.

- Adjusted default parameters to mpileup.  -h has gone from 100 to 500, -Q from 13 to 1, -F from 0.002 to 0.05 and -m from 1 to 2.

- Added a `--indel-bias` mpileup option to skew FN vs FP.  This appears to be better than modifying call `-P` as it's specific to indels.

- Changed the indel scoring algorithm, giving better FP vs FN discrimination.

- Added the short tandem repeat finder from Crumble.  This is used for correcting the indel score.  Put simply, indels in low-complexity regions having a higher bar to pass than those in normal regions.  This is a modest improvement to FN, FP and genotype assignment errors.  (To consider - it may also be worth utilising this on SNPs too.)

- A major refactoring of the extremely long `bcf_call_gap_prep` function.  This was done part way through my work, meaning it's very hard to reorder / squash commits from both sides of that refactor commit.  You may also wish to diff in two stages - up to that commit and beyond that commit, as otherwise it's hard to see what's changed.

- `get_position()` has been rewritten, making it significantly faster on long read data.  Previously it was quite easy for a 1/3rd of the entire CPU time to be spent here.  Now it's negligible.

- Improved calling accuracy for long-read technologies.  This has mainly been tested on PacBio CCS data.  ONT SNP calling is significantly more accurate too, but indels are still a right-off and glacial.  There is more work to be done on the indel caller's use of BAQ, which cripples speed on long-read technology.

- The indel caller now only has one set of `probaln_glocal` parameters instead of two.  I initially removed the second to speed up long-read calling, but it improved it too.  It turned out fortuitously to also improve the Illumina data.  I assume it helped at one point, but on the more modern data sets it served no purpose (was detrimental).

- Added `-X, --config` (although perhaps --platform is better?  Feel free to suggest) to mpileup.  This is simply a collection of parameter values, tuned to work on that specific name.  It could perhaps auto-detect by platform RG fields, but I feel being explicit is likely the safer option.

- Added a `--max-BQ` option to clip (rather than discard) high quality bases.  This is used to limit overly confident platforms from producing unrealistic SNP calls (I'm looking at you CCS!).

- Added a `--delta-BQ` option to permit the neighbouring base quality to be utilised, if lower by a specific delta to the SNP base itself.  This helps CCS calling where often incorrect bases have the low confidence in the next base rather than the one in error.

- Replaced the heuristic for indel allele (REF vs ALT) with a newer, albeit more complex, one.  This fixes #1446 and significantly reduces indel false positive rates.

Some questions to consider:

Should we make `-D` the default?  I think yes, but then how do we define full BAQ?  I guess that can become -D or a similar option.

I have no idea what impact this has on rare variants or on many-sample variant calling, as I don't have truth sets for either so have no way of validating this myself.  I suggest interested parties offer up their services!

This *will* fail tests, as the output differs in so many ways.  However I'm not really sure I'm the best person to judge which differences matter.  I think that's a task for review perhaps.  Obviously if we're happy, we can just generate new expected test results to silence it.


jkbonfield:
Demonstration of the results based on known truth sets (SynDip and GIAB) are in the previous PR here:

https://github.com/samtools/bcftools/pull/1363#issuecomment-802035512

https://github.com/samtools/bcftools/pull/1363#issuecomment-802042033

https://github.com/samtools/bcftools/pull/1363#issuecomment-803929695

I need to redo these again as a few tweaks have happened since, if only to check I haven't created any regressions.  When done I'll post here, but if someone has a specific truth set they want testing then do please let me know.  Note it *must* be a truth set, and not just "does it increase TS/TV".  My experiments there showed it often makes that metric poorer, while getting more results correct, as the previously missing variants were enriched for low-complexity data.

jkbonfield:
As expected I spotted some more minor code tidyups immediately on posting this, so I'll correct and squash again.  My apologies.

jkbonfield:
I tidied up a couple of those commits to remove old debug code or fix some no-longer relevant comments.

I also added a couple new commits.

- Partial-BAQ is now the default behaviour, as it appears to be a win/win scenario and we'd like bcftools out of the box to be fast and effective.  The old full-BAQ has now become the `-D` option.

- Added a `-U, --mwu-u` option to enable the older Mann-Whitney U scoring scale (probabilities between 0 and 1).  This also changes the names of the INFO tags back to the former ones too (BQB vs BQBZ).  This is to permit backwards compatibility, incase people have existing filters.

A question regarding that, should we undo the uni-directional tweak to MWU that I put in for mapping quality, so it is identical?  (Also maybe should it call the old implementation, which still exists but does give slightly different scores I think?)

jkbonfield:
I confirmed I got identical results to my more previous PR, so the mass juggling of commits didn't break things it appears.

I decided to cross-validate my assessments using a different tool, this time RTG's vcf-eval by ploting the produced roc files (without `--squash-ploidy` so genotype errors are taken into consideration).  The VCF has already been filtered with basic rules as per best practices.  I'm just using Chr1 of HG002.

SNPs:
image

image

Bcftools improved, albeit mostly due to fixing the default parameters for low coverage.  On higher coverage the new code is doing much better, beating GATK even.

Indels:
image

image

This is vastly different.  I couldn't zoom up to the more appropriate top-right corner as v1.12 doesn't get remotely close.  It's clear indels are still poor in Bcftools.  Compared to GATK it simply has too many false positives (reducing precision), however the recall is good.  The old code had very poor recall and precision on this data set.

PR #1485, opened 2021-05-10
Fixed typo in doc bcftools.txt

rwk-unil:
Changed "Note that the file must be compressed and index." to
"Note that the file must be compressed and indexed."


pd3:
This is now integrated as a separate commit 6a783f1 due to formatting conflicts. Thanks!

PR #1496, opened 2021-05-26
Support an HTSlib source tree that has a separate build directory

jmarshall:
Followup to PRs samtools/htslib#1277 and samtools/samtools#1427, updating _m4/ax_with_htslib.m4_ here too.

Also updates _INSTALL_'s `--with-htslib` documentation to mention htslib build directories, and lightly expands the `--with-htslib` section for clarity.


pd3:
Thank you

PR #1498, opened 2021-05-27
Fix AD[0] generations for indels

pd3:
Even reads without any evidence for the indel are realigned and
in case their score comes out small (which is the case of high-confidence
reference reads), they are removed from the pileup, reducing AD[0] counts
amongst other. Here we keep all non-indel reads in AD[0].

This change is motivated by the test case test/mpileup/indel-AD.2.*


jkbonfield:
Unfortunately this undoes a lot of the improvements to genotype call accuracy.

An example using GIAB HG002 chr1 at 15x:

Develop branch:
```
SNP          Q>0 /   Q>=20 / Filtered
SNP   TP   97.44 /   95.51 /   95.43
SNP   GT    0.34 /    0.32 /    0.32
SNP   FP    0.71 /    0.27 /    0.25
SNP   FN    2.56 /    4.49 /    4.57

InDel TP   94.35 /   92.14 /   89.94
InDel GT    2.25 /    2.15 /    2.06
InDel FP    1.42 /    1.14 /    0.79
InDel FN    5.65 /    7.86 /   10.06
```

This PR:
```
SNP          Q>0 /   Q>=20 / Filtered
SNP   TP   97.44 /   95.51 /   95.43
SNP   GT    0.34 /    0.32 /    0.32
SNP   FP    0.71 /    0.27 /    0.25
SNP   FN    2.56 /    4.49 /    4.57

InDel TP   93.59 /   91.04 /   88.99
InDel GT    3.61 /    3.44 /    3.31
InDel FP    1.27 /    1.05 /    0.72
InDel FN    6.41 /    8.96 /   11.01
```

Adjusting for FP/FN differences by changing the QUAL filter level, it's possible to get this PR to have higher FP error, higher FN error and higher GT assignment. error.

The original 1.12 code here had:

```
            // This read is not counted as indel. Instead of skipping it, treat it as ref. It is
            // still only an approximation, but gives more accurate AD counts and calls correctly
            // hets instead of alt-homs in some cases (see test/mpileup/indel-AD.1.sam)
            if ( q < bca->min_baseQ ) b = 0, q = (int)bam_get_qual(p->b)[p->qpos];
```

This assumption was plain wrong.  It's basically saying low quality ALT calls are REF, making it very REF biased.  The consequence is it got a huge number of genotype assessments incorrect.  The change makes it even for both, so low quality is filtered out the same.  However it changes depth.  If you want to keep AD high, as it was in 1.12, then we either need to come up with a better way of assessing the quality `q` (preferred IMO), or we need to keep the logic as it currently stands for filtering (the `continue` statement) and add back the missing AD elements.  What other elements would we also fudge this way?

For example:

```
diff --git a/bam2bcf.c b/bam2bcf.c
index a0ed00a..0f5d726 100644
--- a/bam2bcf.c
+++ b/bam2bcf.c
@@ -205,6 +205,8 @@ int bcf_call_glfgen(int _n, const bam_pileup1_t *pl, int ref_base, bcf_callaux_t
 
     // fill the bases array
     double nqual_over_60 = bca->nqual / 60.0;
+    int ADR_ref_missed[4] = {0};
+    int ADF_ref_missed[4] = {0};
     for (i = n = 0; i < _n; ++i) {
         const bam_pileup1_t *p = pl + i;
         int q, b, mapQ, baseQ, is_diff, min_dist, seqQ;
@@ -216,7 +218,15 @@ int bcf_call_glfgen(int _n, const bam_pileup1_t *pl, int ref_base, bcf_callaux_t
         {
             b = p->aux>>16&0x3f;
             seqQ = q = (p->aux & 0xff); // mp2 + builtin indel-bias
-            if (q < bca->min_baseQ) continue;
+            if (q < bca->min_baseQ) {
+                if (!p->indel && b < 4) {
+                    if (bam_is_rev(p->b))
+                        ADR_ref_missed[b]++;
+                    else
+                        ADF_ref_missed[b]++;
+                }
+                continue;
+            }
             if (p->indel == 0 && (q < _n/2 || _n > 20)) {
                 // high quality indel calls without p->indel set aren't
                 // particularly indicative of being a good REF match either,
@@ -321,6 +331,15 @@ int bcf_call_glfgen(int _n, const bam_pileup1_t *pl, int ref_base, bcf_callaux_t
             bca->alt_scl[sc_len]++;
         }
     }
+
+    // Compensate for AD not being counted on low quality REF indel matches.
+    if (r->ADF) {
+        int b;
+        for (b = 0; b < 4; b++) {
+            r->ADR[b] += ADR_ref_missed[b];
+            r->ADF[b] += ADF_ref_missed[b];
+        }
+    }
     r->ori_depth = ori_depth;
     // glfgen
     errmod_cal(bca->e, n, 5, bca->bases, r->p); // calculate PL of each genotype
```

jkbonfield:
With hindsight and looking at examples, I think even fixing it as my idea above is also wrong.

I think we're being too blinkered on keeping the values the same as they were before.  The one example I was given (outside of this PR) the develop branch is actually giving the correct answers.  It may look like a very low AD rate, but it's actually the correct AD rate.

The cause is that in any given alignment column we may have 3 categories when faced with an insertion.  For example:

```
  TAGCC  TATATATATATATATATGGGCCC (ref)
--------------------------------
A TAGCC  TATATATA
B TAGCC  TATATATATATA
C TAGCC  TATATATATATATAT
D TAGCC  TATATATATATATATATatg  (lowercase is soft-clip)
E TAGCCTATATATATATATATATATGGGCCC
F TAGCC  TATATATATATATATATGGGCCC
G TAGCCTATATATATATATATATATGGGCCC
H TAGCC  TATATATATATATATATGGGCCC
I TAGCC  TATATATATATATATATGGGCCC
J TAGCCTATATATATATATATATATGGGCCC
```

Class 1: E, G and J have all span the STR, going ito unique data either side), and have an extra TA in the STR.  D also has an extra TA, but due to ending earlier than the others the aligner chose not to make the TA insertion as it costs more than that ATG match suffix, choosing instead to soft-clip.  So that's 4 for the AT insertion.

Class 2: F, H and I also all span the STR, but they don't have the additional TA and match the reference.

Class 3: A, B and C are aligned against the reference without the additional TA, but as they do not span the STR there is no evidence which of the two alleles they actually belong to.  As a consequence, I would argue that we have 4 ALT, 3 REF, and 3 at qual 0 and filtered out from assessment.

<hr>

Now let's think about a non-heterozygous example for the same region:

```
  TAGCC  TATATATATATATATATGGGCCC (ref)
--------------------------------
A TAGCC  TATATATA
B TAGCC  TATATATATATA
C TAGCC  TATATATATATATAT
D TAGCC  TATATATATATATATATatg  (lowercase is soft-clip)
E TAGCC  TATATATATATATATATatgg  (lowercase is soft-clip)
F TAGCCTATATATATATATATATATGGGCCC
G TAGCCTATATATATATATATATATGGGCCC
H TAGCCTATATATATATATATATATGGGCCC
I TAGCCTATATATATATATATATATGGGCCC
J TAGCCTATATATATATATATATATGGGCCC
```

Naively it looks like 5 with insertion and 5 without, but breaking down into the classifications above we have:

Class 1: 7 ALT (5 called by the aligned, and 2 misaligned).
Class 2: 0 REF
Class 3: 3 that cannot be assigned to REF or ALT as they don't span the STR.

This proposed change would, I think, keep those first 3 reads as-is by not filtering them out due to qual 0, and therefore call a heterozygous insertion instead, leading to the genotype assignment errors.

jkbonfield:
Infact I decided to test it with this exact example (2nd).

[eg.zip](https://github.com/samtools/bcftools/files/6560553/eg.zip)

Gap5 screenshot showing alignments:
image


Bcftools 1.12:

```
$ bcftools112 mpileup -a AD -f mp_ref.fasta mp_hom.sam 2>/dev/null|bcftools112 call -vm - 2>/dev/null|egrep -v '^#'
1	5	.	CTATATATATATATATAT	CTATATATATATATATATAT	102.244.INDEL;IDV=4;IMF=0.444444;DP=9;VDB=0.000273358;SGB=-0.556411;MQ0F=0;AC=1;AN=2;DP4=5,0,4,0;MQ=60	GT:PL:AD	0/1:135,0,192:5,4
```

See `GT 0/1` and `AD 5,4`.  That's the naive assessment.

Bcftools develop branch:

```
$ bcftools mpileup -a AD -f mp_ref.fasta mp_hom.sam 2>/dev/null|bcftools call -vm - 2>/dev/null|egrep -v '^#'
1	5	.	CTATATATATATATATAT	CTATATATATATATATATAT	129.416.INDEL;IDV=4;IMF=0.444444;DP=9;VDB=0.0058656;SGB=-0.556411;RPBZ=-2.57012;SCBZ=-1.34164;FS=0;MQ0F=0;AC=2;AN=2;DP4=0,0,4,0;MQ=60	GT:PL:AD	1/1:159,12,0:0,4
```

GT is now correctly `1/1`.  AD is `0,4` which is too low as it's not accounting for the 2 misaligned reads which match ALT but don't have `p->indel` set.  This isn't ideal, but I don't yet know how to fix that.

Bcftools this PR:

```
$ bcftools mpileup -a AD -f mp_ref.fasta mp_hom.sam 2>/dev/null|bcftools call -vm - 2>/dev/null|egrep -v '^#'
1	5	.	CTATATATATATATATAT	CTATATATATATATATATAT	105.26	.INDEL;IDV=4;IMF=0.444444;DP=9;VDB=0.000273358;SGB=-0.556411;RPBZ=-2.57012;SCBZ=-1.34164;FS=0;MQ0F=0;AC=1;AN=2;DP4=5,0,4,0;MQ=60	GT:PL:AD	0/1:138,0,192:5,4
```

It's back to GT and AD matching 1.12.

Clearly we have some work to do to improve AD further, but I think this PR needs closing and we should figure out a better way.

pd3:
I looked at a random selection of these indel outliers in IGV, see typical examples [here](https://docs.google.com/presentation/d/1mTQYNIGqC0RhDhoWqYn71tJZWg5zvEyQygzNVsimBuM/edit?usp=sharing). The truth set contains some dubious calls, but this does not drive the differences between the development branch and this pull request. The increased FN rate of the initial pull request is driven by reasonable calls which should not have been removed. 

How about adding a new option like this:
```
--ambiguous-reads STR    What to do with ambiguous indel reads
```
where the modes could be:
* drop: when a read aligns with the same quality to REF and ALT allele, it is dropped. This typically happens with reference reads that don't span the entire STR. The ambiguous reads are not used in calling and do not increment the AD counter
* ref: increment the AD[0] counter and consider it as REF allele in calling. This is the initial version of the pull request. This would require adjusting the returned indel quality as for ambiguous reads it would be 0. Lower sensitivity, higher specificity, overcalls hets.
* alt: increment the AD[1] counter and consider it as ALT allele in calling. The inverse of the above.
* ref+drop: increment the AD[0] counter but don't consider the read in calling
* alt+drop: The inverse of the above

jkbonfield:
Every set of parameters who choose will have some "obvious" cases which it doesn't call or it calls incorrectly.  You cannot change parameters based on anecdotal evidence.  It has to be large data set driven.  I don't know what the "called" vs "not called" means in your diagrams though.  Called but with wrong GT?  Nothing called at all (ie absent from VCF file)?    Similarly for false positives.  Why are they false positive?  All 6 of the plots you showed have very clearly deletions.  What's false about the call?  Correct location but wrong sequence?  Correct call but wrong genotype?   Please give some concrete examples.  We need a proper VCF record for the manually curated truth and for the call.

Individual cases are still useful for working out additional layers that can be added to the code.  Eg better post-filtering to remove obvious false positives based on a pattern.  It's a bit trickier on the false negatives though.

Also just because it looks obvious in IGV doesn't mean that there really is obvious evidence for that genotype.  See my examples above.   In IGV you may well jump to the wrong conclusion just looking at a line for the alignment.  I find looking at IGV plots rather counter-intuitive compared to tview or gap5, but if my interpretation of the left 3 of your false negative plots is that they all have the issue I showed above with "REF" matching reads not spanning the STR and likely having invalid alignments.  The ones on the right I'm less sure on.

jkbonfield:
> How about adding a new option like this:
> 
> ```
> --ambiguous-reads STR    What to do with ambiguous indel reads
> ```
> 
> where the modes could be:
> 
>     * drop: when a read aligns with the same quality to REF and ALT allele, it is dropped. This typically happens with reference reads that don't span the entire STR. The ambiguous reads are not used in calling and do not increment the AD counter
>
>    * ref: increment the AD[0] counter and consider it as REF allele in calling. This is the initial version of the pull request. This would require adjusting the returned indel quality as for ambiguous reads it would be 0. Lower sensitivity, higher specificity, overcalls hets.

This is definitely feasible to do, but I'd like to see evidence for these claims.  Can you run your a test on your own data sets, whatever you normally use for truth set analysis with bcftools, showing the FP vs FN tradeoff graphs between dev (drop) and this PR (ref)?  I've shown you (on mattermost) the ones I produced for GIAB HG002 and SynDip chr1 and the evidence isn't conclusive.  There is perhaps a marginal increase in specificity at low depth, but it's not much are a rather poor tradeoff IMO.  I couldn't really see any significant benefit at higher depths but it had a significant detrimental affect on GT assignment.

pd3:
This is now added as an option, the default behavior remains unchanged.

jkbonfield (commented):
To test I tried Chr1 of SynDip at a shallow 15x (for speed).

Develop:

```
SNP   TP   94.24 /   86.81 /   86.76
SNP   GT    0.90 /    0.82 /    0.81
SNP   FP    1.81 /    0.86 /    0.64
SNP   FN    5.76 /   13.19 /   13.24

InDel TP   76.65 /   73.22 /   69.96
InDel GT    5.68 /    5.28 /    4.51
InDel FP    4.42 /    3.28 /    2.00
InDel FN   23.35 /   26.78 /   30.04
```

This PR:
```
SNP   TP   94.24 /   86.81 /   86.76
SNP   GT    0.90 /    0.82 /    0.81
SNP   FP    1.81 /    0.86 /    0.64
SNP   FN    5.76 /   13.19 /   13.24

InDel TP   76.65 /   73.22 /   69.96
InDel GT    5.68 /    5.28 /    4.51
InDel FP    4.42 /    3.28 /    2.00
InDel FN   23.35 /   26.78 /   30.04
```

Those are identicaL, which is what we expect as I didn't enable the new mode.

However running `mpileup --ar incAD` (and `incAD0`) also gave identical VCF files.  I don't know why.  My command line was:

```
##bcftoolsCommand=mpileup --ar incAD0 -f /nfs/srpipe_references/references/Human/1000Genomes_hs37d5/all/fasta/hs37d5.fa CHM1_CHM13_2.15x.chr1.bam
```

Am I missing something?  Do I need to change call too?

pd3:
The calling remains the same in all three cases, only affects the AD values which need to be requested by `-a AD` or `-a INFO/AD`

jkbonfield:
> 
> 
> The calling remains the same in all three cases, only affects the AD values which need to be requested by `-a AD` or `-a INFO/AD`

Ah that makes sense.  I didn't realise they weren't enabled by default.  I'll rerun the tests then.  Maybe on a smaller region.

jkbonfield:
Fixed in 186dc930130a6d936e3c6d2bae30f28f18c3293e, which has an extra minor commit to fix the docs.

PR #1505, opened 2021-06-08
Try to work just with the index file when calculating record statistics

### LINKED ISSUE #1418: Feature request: Allow reading statistics from index file without loading VCF file

valeriuo:
The total number of records can be retrieved directly from an index file and this PR makes it possible to skip opening the variant call file when running `bcftools index --nrecords`. When asking for statistics split by each contig (`bcftools index --stats`), the variant call file is still needed, because the contig names and lengths can only be retrieved from its header.

Also, the user can now specify the index file following the format:
`bcftools index --nrecords variants.vcf.gz##idx##index.vcf.gz.csi`

Fixes #1418

### DIFF
--- a/vcfindex.c
+++ b/vcfindex.c
@@ -37,6 +37,11 @@ DEALINGS IN THE SOFTWARE.  */
 
 #define BCF_LIDX_SHIFT    14
 
+enum {
+    per_contig = 1,
+    total = 2
+};
+
 static void usage(void)
 {
     fprintf(stderr, "\n");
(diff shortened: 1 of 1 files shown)


jmarshall:
This doesn't include any tests. If there were tests exercising the new functionality with data files in which some of the chromosomes have no reads, I expect they would highlight the apparent `hts_idx_nseq()` bug I pointed out in https://github.com/samtools/htslib/pull/1295#issuecomment-855865427.

valeriuo:
The test file collection includes [idx.vcf](https://github.com/samtools/bcftools/blob/develop/test/idx.vcf), with `##contig=<ID=1,length=249250621>` lacking any records. This file is used in [one](https://github.com/samtools/bcftools/blob/7a530fa1a82c19e8aeb02658ae384447ff99535e/test/test.pl#L42) of the tests and the method `hts_idx_nseq` is being called for `bcftools index -n idx.bcf`.

jmarshall:
Good point — there is a test that I guess previously read the variant file too, and its output is unchanged. It turns out that `idx_read_core()` always `kh_init`s every `bidx` entry, so as long as that remains true (and `hts_idx_nseq()` is only called on **loaded** indexes, not freshly created ones) this `hts_idx_nseq()` implementation will return the right value for use with a `hts_idx_get_stat()` loop.

However that means that the `hts_idx_nseq()` documentation is wrong: it's returning the number of chromosomes total, not just those that are covered by reads. And IMHO it would be less misleadingly implemented as just `{ return idx->n; }`.

valeriuo:
> However that means that the `hts_idx_nseq()` documentation is wrong: it's returning the number of chromosomes total, not just those that are covered by reads. And IMHO it would be less misleadingly implemented as just `{ return idx->n; }`.

I see what you mean. I wrote the method `hts_idx_nseq` like that to have it return the same value as `hts_idx_seqnames`, but it was probably not quite right to begin with. It appears `i` is an index in both `seq` and in `hts_idx_t::bidx` (via `hts_idx_get_stat`) and the two don't necessarily match in the number of elements (or their position). I'll amend the code.

freeseek:
With BCFtools 1.11 I was able to extract the per-chromosome statistics, but now it got more complicated.

Before I was able to extract this information from a given `input.bcf.csi` using the following hack:
```
echo -e "##fileformat=VCFv4.2\n#CHROM\tPOS\tID\tREF\tALT\tQUAL\tFILTER\tINFO" | \
  bcftools view -Ob -o tmp.bcf
bcftools index --stats "tmp.bcf##idx##input.bcf.csi"
```

This does not work anymore, as the following new code:
```
        const char *ctg_name = tbx ? seq[tid] : hdr ? bcf_hdr_id2name(hdr, tid) : NULL;
        if ( ctg_name ) {
            bcf_hrec_t *hrec = hdr ? bcf_hdr_get_hrec(hdr, BCF_HL_CTG, "ID", ctg_name, NULL) : NULL;
            int hkey = hrec ? bcf_hrec_find_key(hrec, "length") : -1;
            printf("%s\t%s\t%" PRIu64 "\n", ctg_name, hkey<0?".":hrec->vals[hkey], records);
        }
```
Does not output anything when `ctg_name` is `NULL`.

Now I have to use the following hack to extract the information from the index file (255 is arbitrary here):
```
(echo -e "##fileformat=VCFv4.2"
for i in {0..255}; do echo "##contig=<ID=$i>"; done;
echo -e "#CHROM\tPOS\tID\tREF\tALT\tQUAL\tFILTER\tINFO") | \
  bcftools view -Ob -o tmp.bcf
bcftools index --stats "tmp.bcf##idx##input.bcf.csi"
```

I understand that the choice to suppress the `--stats` functionality for `.csi` index files without a paired `.bcf` file, as the index does not contain the contig ID strings, but it would have been still useful to have some way to extract the per-chromosome information without having to have the whole `.bcf` file available.

I would have preferred the command `bcftools index --stats input.bcf.csi` give the per-chromosome statistics without attempting to read the paired `bcf` file, as it is done with the new code:
```
    /*
     * First, has the user provided an index file? If per contig stats
     * are requested, open the variant file (together with the index file,
     * if provided), since the contig names can only be retrieved from its
     * header. Otherwise, use just the corresponding index file to count
     * the total number of records.
     */
    int len = strlen(fname);
    if ( (fnidx = strstr(fname, HTS_IDX_DELIM)) != NULL ) {
        fntemp = strdup(fname);
        if ( !fntemp ) return 1;
        fntemp[fnidx-fname] = 0;
        fname = fntemp;
        fnidx += strlen(HTS_IDX_DELIM);
    }
    else if ( len>4 && (!strcasecmp(".csi",fname+len-4) || !strcasecmp(".tbi",fname+len-4)) )
    {
        fnidx  = fname;
        fntemp = strdup(fname);
        fname  = fntemp;
        fname[len-4] = 0;
    }

    if ( stats&per_contig )
    {
        fp = hts_open(fname,"r");
        if ( !fp ) {
            fprintf(stderr,"Could not read %s\n", fname);
            ret = 1; goto cleanup;
        }
```
even if this means that no contig names would be provided. It could have provided the per-chromosome statistics using code like this:
```
        printf("%d\t.\t%" PRIu64 "\n", tid, records);
```
or maybe like this:
```
        printf(".\t.\t%" PRIu64 "\n", records);
```

Why is this important? When writing cloud pipelines it can save a lot of time having to localize only the index file without having to localize the whole VCF file and the information in the index file is quite useful to make determinations about memory usage. In my case, I used this hack to extract the number of variants in the largest contig, as this drives the maximum memory consumption for a tool I need to run.

pd3:
This is something I wanted to add but left unfinished. Now supported by https://github.com/samtools/bcftools/commit/02d1723e79c93cfcfb2c3a3470243e601e3888c3

PR #1514, opened 2021-06-23
Summer 2021 Release Copyright Updates.

whitwham:
Some more historical updates in here too sinc this is the first time the
update scripts has been run on bcftools.


pd3:
Thank you!

PR #1581, opened 2021-09-21
Fixes the python plotting routines.

sschmeier:
Small changes to the plotting routines in python that will fix some issues with newer python versions.


pd3:
Thank you

jmarshall:
Note that this was merged to **master**. It might be good to merge/cherry-pick it onto **develop** as well, so that the change is exercised and there are no surprises later.

pd3:
I did not notice, thanks.

jmarshall:
It's a very easy thing to miss :smile: I only noticed because `git fetch` reported that **master** had moved too.

PR #1586, opened 2021-09-27
Make `sort --max-mem` more accurate

### LINKED ISSUE #1576: bcftools keeps getting killed
Hi, 

I'm trying to concatenate and then sort a set of gVCF files.
The particular set of gVCF.gz files sum to 17GB.

After the following concatenation operation

```bash

bcftools \
    concat \
    --naive \
    -f all_raw_vcfs.txt \
    --output-type v \
    -o concatedated_raw.vcf
```

The following sort command keeps getting killed 

```bash
bcftools \
    sort \
    -m ${work_memory}G \
    --temp-dir tm_sort \
    --output-type z \
    -o ${prefix}.vcf.gz \
    concatedated_raw.vcf
```
Even when the machine has 600GB of memory.

So I am wondering
 * what am I doing wrong?
 * what memory footprint should be expected based on input file size?

Thanks!
Steve

------
bcftools version 1.13, compiled from source.

pd3:
Many bcf1_t structures are kept allocated at the same time, each
coming with a small unused memory overhead. Here we allocate
a single large buffer of the maximum allowed size and memcpy the
memory actually used.

Resolves #1576

### DIFF
--- a/vcfsort.c
+++ b/vcfsort.c
@@ -59,6 +59,7 @@ typedef struct _args_t
     int argc, output_type, clevel;
     size_t max_mem, mem;
     bcf1_t **buf;
+    uint8_t *mem_block;
     size_t nbuf, mbuf, nblk;
     blk_t *blk;
 }
@@ -104,8 +105,6 @@ int cmp_bcf_pos(const void *aptr, const void *bptr)
     // This will be called rarely so should not slow the sorting down
     // noticeably.
 
-    if ( !a->unpacked ) bcf_unpack(a, BCF_UN_STR);
-    if ( !b->unpacked ) bcf_unpack(b, BCF_UN_STR);
     int i;
     for (i=0; i<a->n_allele; i++)
     { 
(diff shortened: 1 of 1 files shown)


daviesrob:
This mostly looks OK.  One issue I've noticed is that `args_t::mem` and `args_t::nmem_buf` are both tracking the same value.  They count up in step, which is lucky as both are used at various points in `buf_push()`.  It would be clearer if one of them was removed in favour of the other.

The name `args_t::mem_buf` is quite similar to `args_t::buf`.  It might be better to call it `mem_block` instead.

This does of course make some assumptions about how many parts of the `bcf1_t` struct have been unpacked when the record is copied, and that they are all clean.  I don't suppose this is likely to change any time soon, but it may be worth attempting to future proof it a bit, e.g. by ensuring that the unused pointers in `new_rec->d` are all NULL, and that `new_rec->unpacked` doesn't show them as unpacked.

pd3:
Good suggestions, thank you for reviewing this.

PR #1589, opened 2021-10-07
Add new `bcftools head` command

jmarshall:
Companion PR to samtools/samtools#1517, this adds a new `bcftools head` command for viewing a VCF/BCF file's headers, verbatim. The idea is to have consistent `samtools head` / `bcftools head` commands rather than having to remember which of `-h`/`-H` you want with which `view` command.

`bcftools head` prints the input file's headers and optionally also its first few variant records. (This command always displays the headers as they are in the file — modulo a `##FILTER=<ID=PASS,…>` line — never adding version or command line information itself.)

This PR also adds a `view --with-header` option (equivalent to the default), so that both samtools and bcftools view commands have all three of `--header-only`/`--no-header`/`--with-header` for consistency.
(It also splits the `-h, --header-only` / `-H, --no-header` option usage descriptions into separate lines for clarity.)

And it also reformats the list of commands in _doc/bcftools.txt_ so that the `..` and descriptions line up vertically when viewed with `man bcftools`. (Sadly at the cost of no longer lining up in the _doc/bcftools.txt_ source code.)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,7 +38,7 @@ OBJS     = main.o vcfindex.o tabix.o \
            vcfstats.o vcfisec.o vcfmerge.o vcfquery.o vcffilter.o filter.o vcfsom.o \
            vcfnorm.o vcfgtcheck.o vcfview.o vcfannotate.o vcfroh.o vcfconcat.o \
            vcfcall.o mcall.o vcmp.o gvcf.o reheader.o convert.o vcfconvert.o tsv2vcf.o \
-           vcfcnv.o HMM.o consensus.o ploidy.o bin.o hclust.o version.o \
+           vcfcnv.o vcfhead.o HMM.o consensus.o ploidy.o bin.o hclust.o version.o \
            regidx.o smpl_ilist.o csq.o vcfbuf.o \
            mpileup.o bam2bcf.o bam2bcf_indel.o bam_sample.o \
            vcfsort.o cols.o extsort.o dist.o abuf.o \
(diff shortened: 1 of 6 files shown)


jkbonfield:
The documentation and `--with-header` commits have been merged.  It's likely the head command itself won't land before 1.14.

Fortunately with the CLI adjustment there is a consistent way now to always display headers or never display headers.  There was some debate however over the usefulness of `head` and the desire to not have two different ways of achieving the same goal.  Maybe it'd help if you explain the problem you're solving (that's not already solved with the sanitised view header options).   It's possible we'd be better serviced with a way of specifying `--no-PG` and `--no-version` using a unified option too for further consistency between tools.

Regarding head, I'm curious what your use case is for wanting to display N lines of text.  I can sort of see perhaps where N=1 maybe useful, for version number validation, but htsfile ought to cover that case anyway.  I can't really see a benefit for an arbitrary number of text lines.  If anything, filtering by header line type is far more useful, but that's just grep anyway.

I can see some benefit to reporting M lines of records, but personally I think that's best done in samtools/bcftools `view` command as a file format conversion.  It's a debugging thing really, but I routinely produce subsets of data for experimentation by using htslib's `test_view -N` or `scramble -N` to stop after so many records.  In `head` it's text only, so isn't helpful (to me).

jmarshall:
> If anything, filtering by header line type is far more useful, but that's just grep anyway.

It can be done with `grep`, but it's also useful to build the common queries in so that the tool itself is more expressive. Thanks, `-t, --type TYPE` is a good suggestion for a future enhancement.

PR #1593, opened 2021-10-13
Change the MWU test so MWU 1 or MWU-Z 0 is returned for variance=0.

jkbonfield:
When we only have the one data value, eg 50 alt and 50 ref all with
MQual 60, this gives a variance of 0.  To avoid the division by zero
later, this was simply returned as HUGE_NUM and produces no output
value to filter on.

Instead we now output a probability to explicitly state no bias.

This has no impact when filtering out (view -e), but does when
filtering-in with view -i.


jkbonfield:
Btw I tested it on SynDip (CHM1/13) and the only differences it gave were as expected - a lot more zeros in various *Z format fields.  I also verified that filtering using "view -i" expressions returned more records.

PR #1608, opened 2021-11-05
Minor bcftools merge speedups

jkbonfield:
It's still pretty slow, mainly due to a lot of parsing and info field hackery, but this PR speeds it up by 10-15% by removing most of the dominant function (strcmp).  These come through direct calls (looking for "AN" and "AC" in a tight loop) and excess use of dictionary lookups.

I think there are more dict caching things we could do to avoid more khash usage, but this is a relatively simple start.

### DIFF
--- a/vcfmerge.c
+++ b/vcfmerge.c
@@ -3070,7 +3070,8 @@ void merge_vcf(args_t *args)
 
     int *rid_tab = calloc(args->maux->n, sizeof(*rid_tab));
     if (!rid_tab)
-        return; // NB: no ability to return error code.
+        error("[%s:%d] Could not allocate %ld bytes\n",
+              __FILE__, __LINE__, args->maux->n*sizeof(*rid_tab));
 
     while ( bcf_sr_next_line(args->files) )
     {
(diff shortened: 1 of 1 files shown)


jkbonfield:
Oops sorry, accidentally included my test harness hackery to remove all the tests for code I wasn't changing to speed up the dev cycle.  I'll fix :)

vcfmerge.c
```diff
@@ -3060,13 +3068,17 @@ void merge_vcf(args_t *args)
     args->out_line = bcf_init1();
     args->tmph = kh_init(strdict);
 
+    int *rid_tab = calloc(args->maux->n, sizeof(*rid_tab));
+    if (!rid_tab)
+        return; // NB: no ability to return error code.
```
pd3, line 3073: This must lead to an abort: 
```
if ( !rid_tab )
     error("[%s:%d] Could not allocate %d bytes\n",__FILE__,__LINE__,args->maux->n*sizeof(*rid_tab))
```

pd3:
This is merged now. Thank you

PR #1611, opened 2021-11-10
More flexible read filtering

### LINKED ISSUE #1599: AND logic for mpileup --rf / --ff?
Hi,
There is no documentation of how to do the above. And it also seems to give different results if you set --ff multiple times or try --ff Flag1,Flag2. I think this is a bug.

pd3:
Add new/modify existing mpileup read filtering options to:
    --df, --decl-flags  ..  Skip reads with all of the bits set
    --ef, --excl-flags  ..  Skip reads with any of the bits set
    --if, --incl-flags  ..  Skip reads with none of the bits set
    --nf, --need-flags  ..  Skip reads with some of the bits unset

Resolves #1599


jkbonfield:
My main problem with these names is they fall into the same problem that caused the initial problem.

Some background for those that don't realise - some of these options do different things between samtools and bcftools.  @pd3 - which?  You identified the problem I believe.  I thought it was --rf or --ff, but they looks to be identical.  Precisely which differed and how?

However this comes down to woolly language.  When we say things like "need-flags" or "require-flags", what exactly do we mean?  Are we saying we must have **all** of these flags to keep the record?  Must have **any** (ie at least one) of the flags?  It's just ambiguous, hence apparently having differing implementations of the same thing.

I think our explanations in the short help are more useful than the mnemonics (although I think "some of" should be "any of" for need-flags).  Does "need" mean we need at least one bit matching, or all matching?  It's doubly confusing as we're stating things in the terms of how to accept ("we *need* this"), but it's described in terms of how to reject.  That negative is so easily going to lead to confusion between OR vs AND on bits, which is precisely what this PR was intending to fix.

We seem to be going over this same old ground time and time again, which clearly indicates it's as clear as mud.  So time to be precise and logical.  My alternatives would be:

--df, --skip-all-set .. Skip reads with all of the FLAG bits set
--ef, --skip-any-set .. Skip reads with any of the FLAG bits set
--if, --skip-all-unset .. Skip reads with all of the FLAG bits unset
--nf, --skip-any-unset .. Skip reads with any of the FLAG bits unset

I don't know what shorter options you want.  Are df, ef, if and nf trying to match existing options anywhere?  Where they match (eg rf, ff) we should use them I think for compatibility.  At least some of these match samtools view.

If you prefer to describe in terms of "Keep reads with" (filter-in instead of filter-out) then we can do so, but I'd like them all to be the same style for simplicity.

pd3:
@jkbonfield RE the question what we want, intuitive long options and the short options would ideally be matching the long options. The existing short options will continue to work but will become undocumented and deprecated in future.
Thank you for the suggestions.

jmarshall:
> Some background for those that don't realise - some of these options do different things between samtools and bcftools. pd3 - which? You identified the problem I believe. I thought it was --rf or --ff, but they looks to be identical. Precisely which differed and how?

There is the table of flag filtering options in https://github.com/samtools/samtools/pull/1442#issuecomment-871244137 (which doesn't include any bcftools commands, but one assumes `bcftools mpileup` currently remains the same as `samtools mpileup`; are there other bcftools commands to consider?) — or do you have something else in mind, whereby some of those commands interpreted `--rf` or whichever differently from each other?

Re `--nf, --need-flags`, I believe everything else calls that `--require-flags` or a variant thereof.

What is the proposed `--decl-flags` an abbreviation of?

jkbonfield:
> @jkbonfield RE the question what we want, intuitive long options and the short options would ideally be matching the long options. The existing short options will continue to work but will become undocumented and deprecated in future. Thank you for the suggestions.

As stated I didn't change the short options, so they'd need to be sensible.

Also note my suggestion for names above is predicated by "if we're renaming, then...".  However I'm also still confused as to *why* this is needed at all.  In our meetings you stated it's because the bcftools options don't match the samtools options with the same name, so they needed renaming to avoid confusion, but I can find no evidence of this.  The linked issues #1599 is simply asking for a new option, not a change to all the other options.

Can you clarify the need for renaming, rather than adding 1 new option (which ideally should match the samtools names listed in John's link)?

pd3:
Two options need to be added, not just one, and that makes it a good time to also rename whatever is not named well. For example, it is not intuitive to have `--incl-flags` abbreviated as `--rf`.

jkbonfield:
If my boolean algebra hasn't failed me, the table of meanings for these options are:

Skip when: | (FLAG & _N_) != _N_ | (FLAG & _N_) != 0 | (FLAG & _N_) == 0 | (FLAG & _N_) == _N_
-|-|-|-|-
. | exclude-any-unset | exclude-any-set | exclude-all-unset | exclude-all-set
. | require-all-set | require-all-unset | require-any-set | require-any-unset
existing | -f, --require-flags | -ff, --excl-flags | --rf, --incl-flags |
proposed | --nf, --need-flags | --ef, --excl-flags | --if, --incl-flags | --df, --decl-flags 

I'm assuming the original requirement for this PR is filling out that last column, the `--decl-flags` bit.  That seems like a sensible addition, although I'm not sure what name is best.

I still think if you want to add additional synonyms, they ought to match the more meaningful language descriptions above rather than new terms like "need".  If we really want to go with some non-precise descriptive term, then maybe "reject-flags"?

Our option names come from keep (require),  reject (exclude), keep (include), so the last is likely best termed as a rejection term again.  The middle two (exclude / include) operate on ANY flag matching, while the first and last therefore apply with ALL flags matching.   Hence "reject" as a synonym for exclude-all-set.  But as mentioned above, I strongly dislike the woolly nature of these terms.

mpileup.c
```diff
@@ -1288,12 +1302,30 @@ int main_mpileup(int argc, char *argv[])
         switch (c) {
         case 'x': mplp.flag &= ~MPLP_SMART_OVERLAPS; break;
         case  1 :
-            mplp.rflag_require = bam_str2flag(optarg);
-            if ( mplp.rflag_require<0 ) { fprintf(stderr,"Could not parse --rf %s\n", optarg); return 1; }
+            fprintf(stderr,"The option --rf has been renamed, please use --if instead\n");
+            mplp.rflag_incl = bam_str2flag(optarg);
+            if ( mplp.rflag_incl<0 ) { fprintf(stderr,"Could not parse --if %s\n", optarg); return 1; }
+            break;
+        case  17 :
```
jkbonfield, line 1309: This could be simplified, with case 1 being an fprintf and then immediately falling into the case 17 as 1 and 17 are deliberately identical.

jkbonfield:
> Two options need to be added, not just one, and that makes it a good time to also rename whatever is not named well. For example, it is not intuitive to have `--incl-flags` abbreviated as `--rf`.

Agreed - that's just plain confusing.  I see you retain the old option present too, which is good.  I don't understand why we're changing others though.

pd3:
This thread grows too fast :-)

In my initial reply I just wanted to thank you for your suggestions, which were helpful, and then draft a final proposal later when I put my thinking hat on.

jmarshall:
> For example, it is not intuitive to have --incl-flags abbreviated as --rf.

The existing `--ff` and `--rf` options have gained intuitiveness and familiarity by virtue of being what people have used since 2012 (samtools/samtools@17abb783083d0773b93bc1b61b81c7ab368e3719) and being shared with three other samtools commands. IMHO it would be unfortunate to remove them from this edition of the `mpileup` command.

pd3:
> --df, --skip-all-set .. Skip reads with all of the FLAG bits set
> --ef, --skip-any-set .. Skip reads with any of the FLAG bits set
> --if, --skip-all-unset .. Skip reads with all of the FLAG bits unset
> --nf, --skip-any-unset .. Skip reads with any of the FLAG bits unset

I like the long names. I've been thinking how to make the short names more memorable, does any of these two schemes look better to you? Not ideal, but at least there is some pattern: 
```
--ls, --As, --skip-all-set   .. Skip reads with all of the FLAG bits set
--ns, --as, --skip-any-set   .. Skip reads with any of the FLAG bits set
--lu, --Au, --skip-all-unset .. Skip reads with all of the FLAG bits unset
--nu, --au, --skip-any-unset .. Skip reads with any of the FLAG bits unset
```

jkbonfield:
It's a bit annoying, but really we're in the "I wouldn't start from here" scenario.

If we were starting afresh, absolutely I'd prefer more logical descriptive names that weren't using ambiguous English terms.  However, we are where we are.

If you see the chart in https://github.com/samtools/bcftools/pull/1611#issuecomment-977853409 basically this PR has two changes.

1. It fills out a missing cell in that table.  This was the original issue raised I believe, and it seems like a sensible addition (ideally we'd do the same for other commands, including samtools, but that'd be a separate PR anyway).

2. It renames a bunch of options.

I'm not in favour of part 2.  *If* we were to rename them, then I offered my own choice of alternatives, but I think it's simply too late and will cause confusion rather than resolve it given the speed at which people tend to upgrade software (ie VERY slowly).

pd3:
> 2\. It renames a bunch of options.

I believe we discussed this before, possibly offline. The existing options will continue to function, but will not be advertised going forward. Existing pipelines will continue to work, new users will find more sense.

pd3:
The options have been renamed to
```
        --ls, --skip-all-set    ..  skip reads with all of the FLAG bits set
        --ns, --skip-any-set    ..  skip reads with any of the FLAG bits set
        --lu, --skip-all-unset  ..  skip reads with all of the FLAG bits unset
        --nu, --skip-any-unset  ..  skip reads with any of the FLAG bits unset
```

PR #1621, opened 2021-11-30
scatter variants based on VCF position

freeseek:
I have made some changes to update the functionality of `-r/-R` and `-t/-T` to be on par with those of the `split` plugin and I have fixed a bug in the way `regidx_overlap(...)` was used, so that it is now possible to generate VCFs with non-overlapping variants when the scattered regions are non-overlapping, which was the intended behavior to begin with


pd3:
Thanks!

PR #1624, opened 2021-12-08
Fix vcf-merge when using COLLAPSE_NONE

q623928815:
Fix vcf-merge when using COLLAPSE_NONE:
Only perfect-match triggers the merge event.


pd3:
Can you please provide a test case where this feature does not work with the current version of bcftools (incorrectly) but is fixed with this pull request? The added test case works well, from that perspective it appears no changes are needed.

PR #1648, opened 2022-01-28
Make --indel-bias more sensitive to indels

pd3:
by switching off an existing heuristics which reduces indelQ
and, in the extreme case, has the power to discard an indel
entirely when the reference alignment has low score.

This is a problem for long reads, so newly `--indel-bias 1.01`
is added to 'ont' and 'pacbio-ccs' presets.

Tentative fix to #1459

### DIFF
--- a/bam2bcf.h
+++ b/bam2bcf.h
@@ -1,7 +1,8 @@
 /*  bam2bcf.h -- variant calling.
+                mplp.indel_bias = 1.01;
 
     Copyright (C) 2010-2012 Broad Institute.
-    Copyright (C) 2012-2021 Genome Research Ltd.
+    Copyright (C) 2012-2022 Genome Research Ltd.
 
     Author: Heng Li <lh3@sanger.ac.uk>
 
@@ -99,7 +100,7 @@ typedef struct __bcf_callaux_t {
     uint16_t *bases;        // 5bit: unused, 6:quality, 1:is_rev, 4:2-bit base or indel allele (index to bcf_callaux_t.indel_types)
     errmod_t *e;
     void *rghash;
-    float indel_bias;  // adjusts indel score threshold; lower => call more.
+    float indel_bias_inverted;  // adjusts indel score threshold, 1/--indel-bias, so lower => call more.
 } bcf_callaux_t;
 
 // per-sample values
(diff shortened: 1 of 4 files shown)


jkbonfield:
Unfortunately repurposing an existing option is trashing the intention of the original option.

The first 20 MB of chr1 on SynDip with `--indel-bias 1.0` (the default) gives:

```
SNP   TP   24100 /   23973 /   23947
SNP   FP    1247 /     952 /     612
SNP   GT      83 /      71 /      58
SNP   FN     320 /     447 /     473

InDel TP    3298 /    3244 /    3226
InDel FP     184 /     133 /     113
InDel GT     167 /     158 /     154
InDel FN    1061 /    1115 /    1133
```

With `--indel-bias 1.1`, to favour higher recall rate at the expense of lower precision gives:

```
InDel TP    3304 /    3254 /    3235
InDel FP     192 /     145 /     124
InDel GT     164 /     155 /     150
InDel FN    1055 /    1105 /    1124
```

So we see at filtered Q30 we have +11 FP and -9 FN (plus 4 fewer with the incorrect genotype assignment).  It's a small shift, but it was only a small tweak to the bias value too.

With this PR and `--indel-bias 1.1` we get:

```
InDel TP    3341 /    3304 /    3285
InDel FP     311 /     254 /     213
InDel GT     175 /     169 /     165
InDel FN    1018 /    1055 /    1074
```

Vs the baseline (1.0) that's now +100FP and -59 FN (+11 GT errs).

We need to evaluate the impact of changes on truth sets before proposing them.

pd3:
This is now restructured into a separate option `--no-indelQ-tweaks`. Sadly, it does not help always, still investigating..

jkbonfield:
My big concern with this is whether it'll actually last more than a few months as a half-solution.

It's not working right now on all the data files, so we need to find a better fix, which I'm researching.  It's quite possibly that an improved consensus construction algorithm in the bam2bcf_indel code may make this option completely redundant.

That work isn't going to appear in our next release, so if this option worked more robustly then I'd be inclined to add it anyway.  However the fact it seems somewhere around a 50/50 success rate at rescuing these missing indels, while also having a major jump to false positives, makes me wonder if we should simply reject this in favour of a better implementation down the road.

PR #1760, opened 2022-07-22
fill-tags: Preserve MISSING when copying filter values to INFO and FORMAT

### LINKED ISSUE #1759: Test case `fill-tags  -t 'FORMAT/DP:1=int(smpl_sum(FMT/AD))'` fails on ARM
Running `make test` on an ARM platform produces the following test failure (twice):

```
test_vcf_plugin:
    /home/johnm/bcftools/bcftools +fill-tags --no-version /tmp/…/fill-tags-AD.vcf.gz -- \
        -t 'FORMAT/DP:1=int(smpl_sum(FMT/AD))' | grep -v ^##bcftools_

    The outputs differ:
	/home/johnm/bcftools/test/fill-tags-AD.3.out
	/home/johnm/bcftools/test/fill-tags-AD.3.out.new

9c9
< chr1	10153	.	AC	A	.	.	AD=3,1	AD:DP	10,1:11	.:.	.:.
---
> chr1	10153	.	AC	A	.	.	AD=3,1	AD:DP	10,1:11	.:0	.:0
```

The difference comes down to this float to int conversion:

https://github.com/samtools/bcftools/blob/435d80bfc8a31fd78298f084a6ffd0143166d5fd/plugins/fill-tags.c#L324

For the `.`/`0` fields in question, `src[j]` is a NAN. ~~I am not sure what flavour of NAN, other than that it is neither `bcf_float_missing` nor `bcf_float_vector_end` — so it is probably an actual arithmetic NAN from whatever calculations are involved in `int(smpl_sum(…))`.~~

* On x86-64, `dst[j]` receives the value -2147483648 from casting this NAN, which is printed as MISSING
* On aarch64, `dst[j]` receives the value 0 from casting this NAN, which is printed as a literal zero

This code should handle NANs more carefully; perhaps the `smpl_sum(…)` method needs to special case an empty sample or similar as `bcf_float_missing`, and the _fill-tags.c_ code needs to translate `bcf_float_missing` to `bcf_int32_missing` specially and only use the cast on finite values.

jmarshall:
The array of values returned by the filter may contain `bcf_double_missing` and/or `bcf_double_vector_end` NAN values. Check for these, and convert them both to `bcf_float_missing` / `bcf_int32_missing` as appropriate.

(Thus turning any vector_end values into missing, because the destination array in the range [ncopy,nfill) will be filled with missing, and we prefer to avoid following vector_end with missing.)

Fixes #1759.

### DIFF
--- a/plugins/fill-tags.c
+++ b/plugins/fill-tags.c
@@ -268,6 +268,16 @@ void ftf_destroy(pop_t *pop)
     }
     free(pop->ftf);
 }
+
+#define float_set_from_double(dst, src) \
+    if (bcf_double_is_missing_or_vector_end(src)) bcf_float_set_missing(dst); else (dst) = (src)
+
+static inline int32_t int32_from_double(double src)
+{
+    if (bcf_double_is_missing_or_vector_end(src)) return bcf_int32_missing;
+    else return (int)src;
+}
+
 int ftf_filter_expr(args_t *args, bcf1_t *rec, pop_t *pop, ftf_t *ftf)
 {
     int j,k, nval, nval1;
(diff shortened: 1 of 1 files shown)


pd3:
Thank you

PR #1763, opened 2022-07-27
Remove unused <zlib.h> inclusions [minor]

jmarshall:
Remove remaining references to `bcf_p1_fp_lk`, which has been unused since the undocumented `call -K` option was removed in c4172c5eca9e03113b028676b993c7279e5b1477 in 2013.

After that, neither source file actually uses anything from `<zlib.h>` (_vcfcall.c_ already didn't).

Hence somewhat surprisingly bcftools does not directly use zlib itself (it does indirectly via htslib, of course). So the zlib probing etc could in principle be removed from _configure.ac_ and _Makefile_, but that may not be worth doing at this time.

### DIFF
--- a/prob1.c
+++ b/prob1.c
@@ -30,18 +30,11 @@ THE SOFTWARE.  */
 #include <errno.h>
 #include <assert.h>
 #include <limits.h>
-#include <zlib.h>
 #include "prob1.h"
-
-// #include "kstring.h"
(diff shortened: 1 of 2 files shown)


pd3:
Thank you

jmarshall:
Wow, that was quick! :smile:

PR #1772, opened 2022-08-06
Add `bcftools index --stats --all` option to display all contigs

### LINKED ISSUE #1771: is there a way to extract the list of contigs using bcftools index from index files generated by old versions of tabix?
```
$ wget https://ftp.ncbi.nih.gov/snp/latest_release/VCF/GCF_000001405.39.gz.tbi
$ bcftools index --stats GCF_000001405.39.gz.tbi
```
This does not generate anything. However, the list of contigs could be something useful for the user to be able to retrieve, even if the number of variants per contig is not available. For this specific example, the following code extracts this list:
```
$ zcat GCF_000001405.39.gz.tbi | strings | grep "^NC_\|^NT_\|^NW_" | head
NC_000001.11
NC_000002.12
NC_000003.12
NC_000004.12
NC_000005.10
NC_000006.12
NC_000007.14
NC_000008.11
NC_000009.12
NC_000010.11
```
However, this does not generalize to other index files generated by old versions of tabix

jmarshall:
By default, contigs containing zero records are omitted. Add an option to print these too.

In particular, for old index files generated without metadata pseudo-bins, this is the difference between listing all the contigs and printing nothing. Fixes #1771.

### DIFF
--- a/doc/bcftools.txt
+++ b/doc/bcftools.txt
@@ -1731,6 +1731,10 @@ the CSI first and then the TBI.
     see *<<common_options,Common Options>>*
 
 ==== Stats options:
+*-a, --all*::
+    Used in conjunction with *-s, --stats*, print per contig stats
+    for all contigs, even those with zero records.
+
 *-n, --nrecords*::
     print the number of records based on the CSI or TBI index files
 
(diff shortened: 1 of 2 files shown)


pd3:
Thank you

PR #1812, opened 2022-10-31
Added error checks to plot-roh.pl

### LINKED ISSUE #1788: plot-roh.py - SystemError: tile cannot extend outside image
Hey,
after executing `plot-roh.py inputs/` I get error:
```
plot-roh.py inputs/
/usr/local/bin/plot-roh.py:432: UserWarning: Attempting to set identical left == right == 0 results in singular transformations; automatically expanding.
  ax1.set_xlim(min_x,max_x+xlim*max_x)
Traceback (most recent call last):
  File "/home/damian/anaconda3/envs/aimed4/lib/python3.9/site-packages/PIL/ImageFile.py", line 506, in _save
    fh = fp.fileno()
AttributeError: '_idat' object has no attribute 'fileno'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/bin/plot-roh.py", line 450, in <module>
    plt.savefig(file,dpi=dpi)
  File "/home/damian/anaconda3/envs/aimed4/lib/python3.9/site-packages/matplotlib/pyplot.py", line 859, in savefig
    res = fig.savefig(*args, **kwargs)
  File "/home/damian/anaconda3/envs/aimed4/lib/python3.9/site-packages/matplotlib/figure.py", line 2311, in savefig
    self.canvas.print_figure(fname, **kwargs)
  File "/home/damian/anaconda3/envs/aimed4/lib/python3.9/site-packages/matplotlib/backend_bases.py", line 2210, in print_figure
    result = print_method(
  File "/home/damian/anaconda3/envs/aimed4/lib/python3.9/site-packages/matplotlib/backend_bases.py", line 1639, in wrapper
    return func(*args, **kwargs)
  File "/home/damian/anaconda3/envs/aimed4/lib/python3.9/site-packages/matplotlib/backends/backend_agg.py", line 510, in print_png
    mpl.image.imsave(
  File "/home/damian/anaconda3/envs/aimed4/lib/python3.9/site-packages/matplotlib/image.py", line 1611, in imsave
    image.save(fname, **pil_kwargs)
  File "/home/damian/anaconda3/envs/aimed4/lib/python3.9/site-packages/PIL/Image.py", line 2158, in save
    save_handler(self, fp, filename)
  File "/home/damian/anaconda3/envs/aimed4/lib/python3.9/site-packages/PIL/PngImagePlugin.py", line 1284, in _save
    ImageFile._save(im, _idat(fp, chunk), [("zip", (0, 0) + im.size, 0, rawmode)])
  File "/home/damian/anaconda3/envs/aimed4/lib/python3.9/site-packages/PIL/ImageFile.py", line 514, in _save
    e.setimage(im.im, b)
SystemError: tile cannot extend outside image
```
in the `inputs/` folder I have such `bcftools roh` outputs:
[roh3.txt.gz](https://github.com/samtools/bcftools/files/9528526/roh3.txt.gz)
[roh2.txt.gz](https://github.com/samtools/bcftools/files/9528528/roh2.txt.gz)
[roh1.txt.gz](https://github.com/samtools/bcftools/files/9528529/roh1.txt.gz)

Cheers!

whitwham:
Add an error check and message when the input format is incorrect.  This does not so much fix #1788 as give a meaningful error message rather than a library exception.

Added another error message if the interactive gui cannot be set.

### DIFF
--- a/misc/plot-roh.py
+++ b/misc/plot-roh.py
@@ -361,13 +361,7 @@ off_sep = 0
 dat_rg1 = {}
 
 if not chrs:
-    print("No GT tags found in input file.\n\nPlease try rerunning using run-roh.pl with --roh-args \"-G 30\"")
-    print("\nAlternatively run these commands manually:")
-    print("bcftools annotate --rename-chrs chr-names.txt in.bcf -Ob -o annotated.bcf")
-    print("bcftools roh -G 30 --AF-tag AF1KG annotated.bcf -Orz -o out.file.txt.gz")
-    print("bcftools query -f'GT\\t%CHROM\\t%POS[\\t%SAMPLE\\t%GT]\\n' annotated.bcf | gzip -c >> out.file.txt.gz")
-    print("\nWhere chr-names.txt can be created by:")
-    usage("for i in `seq 1 22`; do printf \"chr$i\\t$i\\n\" >> test_char.txt; done; printf \"chrX\\tX\\n\" >> test_char.txt")
+    usage("No GT lines found in input file. Was it generated using the run-roh.pl script?")
 
 for chr in chrs:
     if chr in dat_rg:
--- a/misc/plot-roh.py
+++ b/misc/plot-roh.py
@@ -361,7 +361,10 @@ off_sep = 0
 dat_rg1 = {}
 
 if not chrs:
-    usage("No GT lines found in input file. Was it generated using the run-roh.pl script?")
+    usage("\nNo GT lines found in input file. Was it generated using the run-roh.pl script?\n"
+          "This script is unable to process the raw output of \"bcftools roh\",\n"
+          "the output of the following query may need to be added to the roh result.\n\n"
+          "bcftools query -f'GT\\t%CHROM\\t%POS[\\t%SAMPLE\\t%GT]\\n' in.bcf\n")
 
 for chr in chrs:
     if chr in dat_rg:
--- a/misc/plot-roh.py
+++ b/misc/plot-roh.py
@@ -133,14 +133,21 @@ if interactive==False:
     import matplotlib.pyplot as plt
     import matplotlib.patches as patches
 else:
+    gui_set = False;
+
     for gui in ['TKAgg','GTKAgg','Qt4Agg','WXAgg','MacOSX']:
         try:
             mpl.use(gui,warn=False, force=True)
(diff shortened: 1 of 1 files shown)


misc/plot-roh.py
```diff
@@ -352,6 +359,16 @@ def parse_samples(fname,highlight):
 off = 0
 off_sep = 0
 dat_rg1 = {}
+
+if not chrs:
+    print("No GT tags found in input file.\n\nPlease try rerunning using run-roh.pl with --roh-args \"-G 30\"")
```
pd3, line 364: This is misleading, the -G30 option is for something else. It's for when the run-roh.pl script fails because there are no FORMAT/PL tags in the VCF.

When there are no GT lines present in the input to plot-roh.py, that would be caused by the user trying to feed in raw output of `bcftools roh`. So the suggestion should be something like "there are no GT lines in the input file, was it generated using the run-roh.pl script?"

misc/plot-roh.py
```diff
@@ -352,6 +359,16 @@ def parse_samples(fname,highlight):
...
+    print("\nAlternatively run these commands manually:")
+    print("bcftools annotate --rename-chrs chr-names.txt in.bcf -Ob -o annotated.bcf")
```
pd3, line 366: This is not something to be applied blindly, even though run-roh.pl does it (the script is a bit simplistic in this sense). This would be used only when the chromosome naming convention is different in the genetic map file or in the AF file. However, as it's not used anywhere in the commands below, it does not need to be listed here. Only the `query` command is essential - whichever way the user produced the roh output, now only the GT lines need to be added.

whitwham: So, should I just take this bit out?

misc/plot-roh.py
whitwham: Okay, I misunderstood.  I will change it.

misc/plot-roh.py
pd3: Yeap :+1:

misc/plot-roh.py
whitwham: Done ...

misc/plot-roh.py
whitwham: ... and done.

pd3:
Thank you for the pull request and testing.

PR #1838, opened 2022-12-19
concat: don't always error when setting --verbose

kdm9:
Currently giving `bcftools concat --verbose` with any argument errors out due to a missing conditional. This check implements the conditional which the error message describes.


pd3:
Thank you

PR #1879, opened 2023-03-06
add missing M option in getopt call

mbhall88:
When running `filter` with the `-M` option I get the following error message 

```
filter: invalid option -- 'M'
```

I noticed it wasn't present in the list of chars passed to `getopt_long`


pd3:
Thank you

PR #1881, opened 2023-03-07
Add --write-index to bcftools view

jkbonfield:
Note there are some differences to output due to htslib not calling `hts_idx_amend_last`, but adding this to htslib/vcf.c cures that (for VCF.gz at least).  Practically speaking, the change is irrelevant as it's simply two different ways of specifying the same file byte offset.  (One past end of last block vs zero bytes into next block)

### DIFF
--- a/vcfview.c
+++ b/vcfview.c
@@ -76,6 +76,8 @@ typedef struct _args_t
     char *include_types, *exclude_types;
     int include, exclude;
     int record_cmd_line;
+    char *index_fn;
+    int write_index;
     htsFile *out;
 }
 args_t;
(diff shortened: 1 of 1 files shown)


pd3:
Thank you. I will add this to other bcftools commands as well.

jkbonfield:
You may have wanted to wait for the tests to finish.  While it passed make check on my local machine, apparently it's broken on AppVeyor.

pd3:
That's okay, it fails on something else, unrelated to indexing.

pd3:
I ran a quick test. For CSI creation it works, but for TBI it crashes inside htslib. Is this the correct format to enforce custom index `file.vcf.gz##idx##file.vcf.gz.tbi`?
```
==26704== Invalid read of size 8
==26704==    at 0x1F3EAE: bgzf_flush (bgzf.c:1940)
==26704==    by 0x21BB97: sam_idx_save (sam.c:1070)
==26704==    by 0x166292: main_vcfview (vcfview.c:808)
==26704==    by 0x6A45C86: (below main) (libc-start.c:310)
==26704==  Address 0xc3a62a8 is 8 bytes after a block of size 64 alloc'd
==26704==    at 0x4C31B0F: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==26704==    by 0x1FB60E: hfile_init (hfile.c:107)
==26704==    by 0x1FC078: hopen_fd (hfile.c:624)
==26704==    by 0x1FD36C: hopen (hfile.c:1276)
==26704==    by 0x2079CF: hts_open_format (hts.c:884)
==26704==    by 0x165F72: init_data (vcfview.c:228)
==26704==    by 0x165F72: main_vcfview (vcfview.c:783)
==26704==    by 0x6A45C86: (below main) (libc-start.c:310)
```

jkbonfield:
It should be the correct format yes.  I have been using this in my testing.  I haven't built it with address sanitizer enabled or under valgrind though

jkbonfield:
Although we probably shouldn't be needing to use the `##idx##` notation generally.  It ought to pick the most appropriate index format for the file type so 9 times out of 10 it's doing the right thing.

pd3:
I'll send you the file offline. This is the command line
```
bcftools view rmme.bcf -o rmmex.vcf.gz##idx##rmmex.vcf.gz.csi --write-index
```

jkbonfield:
Ah, this appears to demonstrate a bug in bcftools filename parsing.

`-o a.vcf.gz` writes gzipped vcf, while `-o a.vcf.gz##idx##a.vcf.gz.csi` writes uncompressed VCF (and an index that cannot possibly work) as the filename doesn't end in .gz.

Samtools strips off the ##idx## bit before checking, but apparently bcftools doesn't do that.

jkbonfield:
See https://github.com/samtools/htslib/pull/1581 for the robustness fix to this.  We weren't validating the file we attempted to write an index for was bgzf compressed.  This is an API usage error, but it's good form to check and reject.

Plus https://github.com/samtools/bcftools/pull/1882 here fixes bcftools command line parsing to auto-detect the format better when using the `##idx##` notation, which fixes the reason we were triggering the htslib problem.

pd3:
Uh, I fixed this in the meantime in my branch and also adding this to all the other commands. Thanks!

jkbonfield:
Ah ok, feel free to close #1882 then as you've surpassed it already.

PR #1882, opened 2023-03-07
Vcf format detection

jkbonfield:
- Make a.vcf.gz##idx##a.vcf.gz.csi do the right thing.

- Flush & close the VCF/BCF file if index writing failed, so we're in a known-good state.


pd3:
This is now fixed and the support for all remaining commands added in 9b352b7c5431b5cbb33dd5141864c389a1c8e0a0

PR #1903, opened 2023-04-06
Add error checking to gvcf_write_block

### LINKED ISSUE #1900: Failure in bcftools merge
On Ubuntu 18.04.6 LTS (bionic) the test harness is getting a segmentation fault in the merge test.  I've fixed the segv, but it still fails.

It can be reproduced with:

```
$ bgzip < test/merge.gvcf.10.a.vcf > _a.vcf.gz
$ bgzip < test/merge.gvcf.10.b.vcf > _b.vcf.gz
$ bcftools index _a.vcf.gz
$ bcftools index _b.vcf.gz
$ ./bcftools merge --no-version -g test/merge.gvcf.10.fa _a.vcf.gz _b.vcf.gz
[E::faidx_adjust_position] The sequence "chr2" was not found
Segmentation fault
```

The segv can be fixed with this:

```
diff --git a/vcfmerge.c b/vcfmerge.c
index 5bd7ad6a..20b71ac5 100644
--- a/vcfmerge.c
+++ b/vcfmerge.c
@@ -2564,6 +2564,9 @@ void gvcf_write_block(args_t *args, int start, int end)
     {
         int slen  = 0;
         char *seq = faidx_fetch_seq(args->gvcf_fai,maux->chr,out->pos,out->pos,&slen);
+        if (!seq)
+            exit(1); // faidx_fetch_seq has already reported the error.
+
         if (slen)
         {
             out->d.allele[0][0] = seq[0];
```

It's a foul way of reporting the error, but `gvcf_write_block` is void so there's no way of reporting back errors, and besides the uses of `error()` everywhere else ends in an `exit(-1)` call.  (Note -1 isn't a good exit code as they're meant to be positive I think)

Anyway, that just stops the untidy crash.  It doesn't stop the failure.  I don't understand why it's broken.  The data in question has:

a.vcf.gz
```
##fileformat=VCFv4.2
##FILTER=<ID=PASS,Description="All filters passed">
##contig=<ID=chr1,length=248956422>
##contig=<ID=chr2,length=248956422>
##INFO=<ID=END,Number=1,Type=Integer,Description="Stop position of the interval">
##FORMAT=<ID=GT,Number=1,Type=String,Description="Genotype">
#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	a
chr1	1	.	A	<*>	.	.	END=2	GT	0/0
chr2	1	.	A	<*>	.	.	END=2	GT	0/0
chr2	3	.	G	C,<*>	.	.	.	GT	0/1
chr2	4	.	T	<*>	.	.	END=6	GT	0/0
```

and b.vcf.gz
```
##fileformat=VCFv4.2
##FILTER=<ID=PASS,Description="All filters passed">
##contig=<ID=chr1,length=248956422>
##contig=<ID=chr2,length=248956422>
##INFO=<ID=END,Number=1,Type=Integer,Description="Stop position of the interval">
##FORMAT=<ID=GT,Number=1,Type=String,Description="Genotype">
#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	b
chr1	1	.	A	C	.	.	.	GT	0/1
chr2	3	.	G	A,<*>	.	.	.	GT	0/1
chr2	4	.	T	<*>	.	.	END=8	GT	0/0
```

Both have a chr2, so the warning message is wrong.  Note on the CI tests this passes, so it's OS specific.

jkbonfield:
This avoids a segmentation fault when bcftools merge uses an incomplete fasta .fai file.

Fixes #1900

For some definition of "fixes".  It now exists cleanly instead of simply dumping core due to a null-pointer exception.  The cause of the error in 1900 though was incompatible .fa and .fai files, caused by "git pull" on an existing clone.  To consider: make check could remove all the auto-generated .fai files to avoid such breakages.

### DIFF
--- a/vcfmerge.c
+++ b/vcfmerge.c
@@ -2564,6 +2564,9 @@ void gvcf_write_block(args_t *args, int start, int end)
     {
         int slen  = 0;
         char *seq = faidx_fetch_seq(args->gvcf_fai,maux->chr,out->pos,out->pos,&slen);
+        if (!seq)
+            exit(1); // faidx_fetch_seq has already reported the error.
+
         if (slen)
         {
             out->d.allele[0][0] = seq[0];


vcfmerge.c
```diff
@@ -2564,6 +2564,9 @@ void gvcf_write_block(args_t *args, int start, int end)
     {
         int slen  = 0;
         char *seq = faidx_fetch_seq(args->gvcf_fai,maux->chr,out->pos,out->pos,&slen);
+        if (!seq)
+            exit(1); // faidx_fetch_seq has already reported the error.
```
pd3, line 2568: Better to use `error()` to explain what happened.

jkbonfield: It doesn't need it as faidx_fetch_seq has already used `hts_log_error` to explain the problem.  See the text in #1900:

```
[E::faidx_adjust_position] The sequence "chr2" was not found
Segmentation fault
```

It's the second line this is fixing. :-)

PR #1914, opened 2023-04-24
Stop treating multiple CIGAR match operations as indels.

whitwham:
Multiple matching (or sequence (mis)matching)) are treated as containing indels.  This can lead to realignment not being called.

Fixes samtools/htslib#1597 in conjunction with samtools/htslib#1607.


pd3:
Thank you

PR #1920, opened 2023-05-16
bcftools +fill-tag: include F_MISSING in --tag all

kdm9:
Hello,

A simple patch that ensures `INFO/F_MISSING` is calculated when doing `bcftools +fill-tags in.vcf.gz -- --tags all`. All other "standard" tags are, and it confused me that a tag explicitly listed in the man page was not included when using `--tags all`.

Best,
K


pd3:
Updated tests and merged, thank you.

PR #1941, opened 2023-06-09
 average(FORMAT/GQ) for vcfstats

lindenb:
Hi all,

This is a `preliminary` pull request.

I wanted to get the average **genotype** **quality** for my variants.

this small PR adds a 15th column containing the average `GQ`  to the `PSC` output of `vcfstats`.

I didn't change the expected test outputs but I will if you tell me that the idea PR is OK 

Thanks,
Pierre

### DIFF
--- a/misc/plot-vcfstats
+++ b/misc/plot-vcfstats
@@ -212,7 +212,7 @@ sub parse_params
             {
                 id=>'PSC',
                 header=>'PSC, Per-sample counts',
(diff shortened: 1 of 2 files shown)


pd3:
Thanks for the pull request. I am thinking if it should be made optional and if it would be useful to make into something more general. We already have average depth, maybe there are other fields that would occasionally be useful to collect.

PR #1968, opened 2023-07-29
Move most gff.h includes to source code gff.c; update Makefile dependencies [minor]

jmarshall:
Various dependency fixes, and use `PRIu32` etc in _gff.c_. Mostly minor.


pd3:
Thank you

PR #1970, opened 2023-08-01
Stop make check from running tests twice

daviesrob:
`check` runs `test-plugins` via a rule on [Makefile line 203](https://github.com/daviesrob/bcftools/blob/5940e5c61dac80d45bdb554af4b6b939a11a5873/Makefile#L203), so we don't want it to run `test-no-plugins` as well.


pd3:
Thank you

PR #2040, opened 2023-11-20
make USE_GPL=1 clean all fails because GSL_LIBS already defined

tobiasrausch:
Hi,

Small fix for:

`make USE_GPL=1 clean all`

`GSL_LIBS ?= -lgsl -lcblas` didn't work for me because GSL_LIBS was already defined in the Makefile beforehand.

Best, Tobias

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -142,7 +142,7 @@ print-version:
 ifdef USE_GPL
     main.o : EXTRA_CPPFLAGS += -DUSE_GPL
     OBJS += polysomy.o peakfit.o
-    GSL_LIBS ?= -lgsl -lcblas
+    GSL_LIBS += -lgsl -lcblas
 endif
 
 print-%:
--- a/Makefile
+++ b/Makefile
@@ -142,7 +142,9 @@ print-version:
 ifdef USE_GPL
     main.o : EXTRA_CPPFLAGS += -DUSE_GPL
     OBJS += polysomy.o peakfit.o
-    GSL_LIBS += -lgsl -lcblas
+    #ifndef GSL_LIBS
+        GSL_LIBS += -lgsl -lcblas
+    #endif
 endif
 
 print-%:
--- a/Makefile
+++ b/Makefile
@@ -142,9 +142,9 @@ print-version:
 ifdef USE_GPL
     main.o : EXTRA_CPPFLAGS += -DUSE_GPL
     OBJS += polysomy.o peakfit.o
-    #ifndef GSL_LIBS
+    ifndef GSL_LIBS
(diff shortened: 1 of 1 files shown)


daviesrob:
Sorry to be a while getting on to this.

The proposed solution isn't quite correct as it doesn't account for values set by `configure`.  Currently, enabling libgsl through `configure` does work, as can be seen using `make print-GSL_LIBS`:

```
$ ./configure --enable-libgsl
[ ... output snipped ... ]
checking for gsl/gsl_version.h... yes
checking for library containing cblas_dgemm... -lcblas
checking for gsl_multifit_gradient in -lgsl... yes
checking for library containing regcomp... none required
configure: creating ./config.status
config.status: creating config.mk
[ ... output snipped ... ]

$ make print-GSL_LIBS
GSL_LIBS=-lgsl -lcblas
```

As noted, either not running `configure`, or running it without `--enable-libgsl` and then trying to enable libgsl using `make USE_GPL=1` doesn't work:

```
$ ./configure
[ ... output omitted ... ]

$ make print-GSL_LIBS
GSL_LIBS=

$ make USE_GPL=1 print-GSL_LIBS
GSL_LIBS=
```

The proposed solution fixes that, but when you run `configure` it adds some unwanted entries to `GSL_LIBS`:

```
$ ./configure --enable-libgsl
[ ... output omitted ... ]

$ make print-GSL_LIBS
GSL_LIBS=-lgsl -lcblas -lgsl -lcblas
```

In this case it's harmless, but it may clash with the decisions made by `configure` in some cases.

I think a solution that gives the intended behaviour might be:

```
ifdef USE_GPL
    main.o : EXTRA_CPPFLAGS += -DUSE_GPL
    OBJS += polysomy.o peakfit.o
    #ifndef GSL_LIBS                                                             
        GSL_LIBS += -lgsl -lcblas
    #endif                                                                       
endif
```

as `ifndef` matches both unset and empty (before expansion) variables.  With that I get:

```
$ ./configure
[ ... output omitted ... ]
$ make print-GSL_LIBS
GSL_LIBS=
$ make USE_GPL=1 print-GSL_LIBS
GSL_LIBS=-lgsl -lcblas

$ ./configure --enable-libgsl
[ ... output omitted ... ]
$ make print-GSL_LIBS
GSL_LIBS=-lgsl -lcblas
$ make USE_GPL=1 print-GSL_LIBS
GSL_LIBS=-lgsl -lcblas
```
... which is what we want to happen.

tobiasrausch:
Thanks. I missed the configure part (sorry) but your solution seems to work fine.

pd3:
Thank you both!

PR #2062, opened 2023-12-20
Improve docs on --soft-filter argument.

SpikyClip:
I experienced some confusion trying to figure out why my filter tags were inversely being applied when using `--soft-filter`. I was under the initial impression that it applied `STRING` to records that meet the filter expression, just like how it applies `STRING` to any records that are in `--mask-file`.

In hindsight it makes sense that it is applied to things that fail to meet the expression, but I thought the documentation on --soft-filter could be improved to consider this discrepancy in behaviour between `--mask-file` and `--include/--exclude`. Especially considering it may not be immediately obvious as no errors are thrown.


pd3:
Thank you

PR #2071, opened 2024-01-06
transcript select for mane and pick in split-vep

barslmn:
I copy-pasted from the CANONICAL selection.
Only difference is with matching the mane transcript where there is additional function call to strncpy for matching "NM".

### DIFF
--- a/plugins/split-vep.c
+++ b/plugins/split-vep.c
@@ -51,6 +51,8 @@
 #define SELECT_TR_ALL       0
 #define SELECT_TR_WORST     1
 #define SELECT_TR_PRIMARY   2
+#define SELECT_TR_MANE      3
+#define SELECT_TR_PICK      4
 #define SELECT_CSQ_ANY      -1
 
 #define GENES_RESTRICT   0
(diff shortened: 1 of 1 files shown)


pd3:
Thank you for the pull request. Before it can be considered, the test suite must pass and tests for this new functionality added.

barslmn:
Hi, tests are passing. 

This functionality does depend on the vep command being used having the `--pick` parameter. As well as the reference genome since hg19 wont have the mane select.

pd3:
Thank you, it looks good at a brief glance. It will still need a test case before it can be merged.

barslmn:
I can supply test cases and write the tests. Would you direct me where do I put these tests or any guidelines? I only see the data files under test/ dir.

pd3:
Actually, don't worry about it, I will add the tests myself. I just got a similar request and realized it will be better to turn this into something more general, based on regular expressions. I will do that, adding your 'pick' and 'mane' functionality as presets.

barslmn:
Thanks a lot

pd3:
This has been now implemented in https://github.com/samtools/bcftools/commit/944fc9339f441f25c879170aaa7dc530e55d7d9b

PR #2135, opened 2024-03-21
Add mpileup_bench/run_multi.sh for multi-sample evaluations

jkbonfield:
This doesn't change the test harness at all.  It's simply extra documentation and scripts to aid with large scale evaluation of mpileup changes.


jkbonfield:
Quick mod inc - my guesswork on the filenames after split was wrong.

PR #2136, opened 2024-03-22
Make --write-index=tbi work.

### LINKED ISSUE #2008: [Feature Request] Allow --write-index to choose tbi index output
Hi,

This is a small feature request that `--write-index` take an optional parameter to allow it to write `.tbi` index files if desired. Something like `--write-index tbi` or `--write-index t` would be great to have since currently the flag just outputs `.csi` indices. I looked through the code a little and it wasn't obvious how this flag gets parsed and used, but if anyone could offer some guidance on where the code might have to be updated to add this, I could take a look at adding the feature.

jkbonfield:
This adds a new function `init_index2` with an additional index format field, which is set by another new function `write_index_parse`.  I could have modified the existing `init_index` function, but I was unsure if this is meant to be a public API or not.  Eg do people write their own plugins and use this?  If not, we could just rename it back.

The PR makes `--write-index` take an optional `--write-index[=FMT]` syntax, eg `--write-index=tbi`.  As this is a long option, the optional argument is easy to define.  (They don't work well on short opts, being space concious.)  The default of not specified =FMT is whatever the tool did before.  This is almost always CSI, except for isec which is TBI if VCF.  I don't know why that single command has a different default, but I didn't want to second guess the logic so I left it unchanged.

I tested all subcommands capable of writing indices, including plugins (although just fill-tags for the generic writing in plugin.c as it's shared by many).  I also fixed a number of bugs, not all index related.

- `bcftools +scatter -n 100 --write-index` only generated an index for ther last file.
- `bcftools isec --write-index` only worked when outputting a single file.  It now honours the write-index when running in directory mode too, but as mentioned above for some reason single-file isec was always tbi so I kept this behaviour intact.
- The malformed `bcftools filter -i 'type=INDEL'` no longer core dumps.  (The correct syntax is type="INDEL")

Fixes #2008

### DIFF
--- a/csq.c
+++ b/csq.c
@@ -554,7 +554,9 @@ void init_data(args_t *args)
         if ( args->hdr_nsmpl )
             bcf_hdr_printf(args->hdr,"##FORMAT=<ID=%s,Number=.,Type=Integer,Description=\"Bitmask of indexes to INFO/BCSQ, with interleaved first/second haplotype. Use \\\"bcftools query -f'[%%CHROM\\t%%POS\\t%%SAMPLE\\t%%TBCSQ\\n]'\\\" to translate.\">",args->bcsq_tag);
         if ( bcf_hdr_write(args->out_fh, args->hdr)!=0 ) error("[%s] Error: cannot write the header to %s\n", __func__,args->output_fname?args->output_fname:"standard output");
-        if ( args->write_index && init_index(args->out_fh,args->hdr,args->output_fname,&args->index_fn)<0 ) error("Error: failed to initialise index for %s\n",args->output_fname);
+        if ( init_index2(args->out_fh,args->hdr,args->output_fname,
+                         &args->index_fn, args->write_index) < 0 )
+            error("Error: failed to initialise index for %s\n",args->output_fname);
     }
     if ( args->verbosity > 0 ) fprintf(stderr,"Calling...\n");
 }
@@ -3328,7 +3330,7 @@ static const char *usage(void)
         "       --targets-overlap 0|1|2       Include if POS in the region (0), record overlaps (1), variant overlaps (2) [0]\n"
         "       --threads INT                 Use multithreading with <int> worker threads [0]\n"
         "   -v, --verbose INT                 Verbosity level 0-2 [1]\n"
-        "       --write-index                 Automatically index the output files [off]\n"
+        "       --write-index[=FMT]           Automatically index the output files [off]\n"
         "\n"
         "Example:\n"
         "   bcftools csq -f hs37d5.fa -g Homo_sapiens.GRCh37.82.gff3.gz in.vcf\n"
@@ -3379,7 +3381,7 @@ int main_csq(int argc, char *argv[])
         {"targets-file",1,0,'T'},
         {"targets-overlap",required_argument,NULL,5},
         {"no-version",no_argument,NULL,3},
-        {"write-index",no_argument,NULL,6},
+        {"write-index",optional_argument,NULL,6},
         {"dump-gff",required_argument,NULL,7},
         {"unify-chr-names",required_argument,NULL,8},
         {0,0,0,0}
@@ -3470,7 +3472,10 @@ int main_csq(int argc, char *argv[])
                 targets_overlap = parse_overlap_option(optarg);
                 if ( targets_overlap < 0 ) error("Could not parse: --targets-overlap %s\n",optarg);
                 break;
-            case  6 : args->write_index = 1; break;
+            case  6 :
+                if (!(args->write_index = write_index_parse(optarg)))
+                    error("Unsupported index format '%s'\n", optarg);
+                break;
             case  7 : args->dump_gff = optarg; break;
             case  8 :
                 if ( !strcmp(optarg,"0") ) args->unify_chr_names = 0;
--- a/csq.c
+++ b/csq.c
@@ -3495,7 +3495,7 @@ int main_csq(int argc, char *argv[])
     }
     else fname = argv[optind];
     if ( argc - optind>1 ) error("%s", usage());
-    if ( !args->fa_fname ) error("Missing the --fa-ref option\n");
+    if ( !args->fa_fname ) error("Missing the --fasta-ref option\n");
     if ( !args->gff_fname ) error("Missing the --gff option\n");
     args->sr = bcf_sr_init();
     if ( targets_list )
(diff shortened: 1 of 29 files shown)


csq.c
```diff
@@ -3328,7 +3330,7 @@ static const char *usage(void)
         "       --targets-overlap 0|1|2       Include if POS in the region (0), record overlaps (1), variant overlaps (2) [0]\n"
         "       --threads INT                 Use multithreading with <int> worker threads [0]\n"
         "   -v, --verbose INT                 Verbosity level 0-2 [1]\n"
-        "       --write-index                 Automatically index the output files [off]\n"
+        "       --write-index[=FMT]           Automatically index the output files [off]\n"
```
pd3, line 3333: Please add an example, something like
```
--write-index[=FMT]           Automatically index the output files, FMT={csi|tbi} [off]
```

pd3:
Thank you for this! Can you update the documentation as well please, doc/bcftools.txt?

csq.c
pd3: Maybe a short option `-W, --write-index` could be added at the same time as well? (Requested here https://github.com/samtools/bcftools/issues/2139)

jkbonfield: I'm wary of padding the usage lines even longer.  They already don't fit on a standard 80 column window.  I think adding to documentation is fine though.

jkbonfield: Short options are a nightmare when it comes to optional arguments.  I mentioned this in my comment before.

I think if we want to permit -W, then we should change how we handle the optional bit as it's just a recipe for confused users frankly.

A case in point, `-W tbi` is an error, `-W=tbi` is an error, but `-Wtbi` works.  With long options it's obvious.

We could perhaps change the parsing so "=tbi" and "tbi" are both recognised as index formats, along with "=csi" and "csi", but it is a non-standard approach to optional arguments.

jkbonfield:
So this change would, for example, allow `-W=tbi` to work as a short option.  It's not really the way short options are meant to be used with optional arguments however, but I feel it's less likely to be a source of confusion and questions than going with the official nomenclature of `-Wtbi` (which still works).  Thoughts?

```
diff --git a/vcfview.c b/vcfview.c
index c53ca1d0..43845af7 100644
--- a/vcfview.c
+++ b/vcfview.c
@@ -550,7 +550,7 @@ static void usage(args_t *args)
     fprintf(stderr, "    -u/U, --uncalled/--exclude-uncalled    Select/exclude sites without a called genotype\n");
     fprintf(stderr, "    -v/V, --types/--exclude-types LIST     Select/exclude comma-separated list of variant types: snps,indels,mnps,ref,bnd,other [null]\n");
     fprintf(stderr, "    -x/X, --private/--exclude-private      Select/exclude sites where the non-reference alleles are exclusive (private) to the subset samples\n");
-    fprintf(stderr, "          --write-index[=FMT]              Automatically index the output files [off]\n");
+    fprintf(stderr, "    -W,   --write-index[=FMT]              Automatically index the output files [off]\n");
     fprintf(stderr, "\n");
     exit(1);
 }
@@ -617,11 +617,11 @@ int main_vcfview(int argc, char *argv[])
         {"phased",no_argument,NULL,'p'},
         {"exclude-phased",no_argument,NULL,'P'},
         {"no-version",no_argument,NULL,8},
-        {"write-index",optional_argument,NULL,10},
+        {"write-index",optional_argument,NULL,'W'},
         {NULL,0,NULL,0}
     };
     char *tmp;
-    while ((c = getopt_long(argc, argv, "l:t:T:r:R:o:O:s:S:Gf:knv:V:m:M:aAuUhHc:C:Ii:e:xXpPq:Q:g:",loptions,NULL)) >= 0)
+    while ((c = getopt_long(argc, argv, "l:t:T:r:R:o:O:s:S:Gf:knv:V:m:M:aAuUhHc:C:Ii:e:xXpPq:Q:g:W::",loptions,NULL)) >= 0)
     {
         char allele_type[9] = "nref";
         switch (c)
@@ -750,7 +750,7 @@ int main_vcfview(int argc, char *argv[])
                 break;
             case  9 : args->n_threads = strtol(optarg, 0, 0); break;
             case  8 : args->record_cmd_line = 0; break;
-            case 10 :
+            case 'W' :
                 if (!(args->write_index = write_index_parse(optarg)))
                     error("Unsupported index format '%s'\n", optarg);
                 break;
diff --git a/version.c b/version.c
index ea112f60..cf2c6d51 100644
--- a/version.c
+++ b/version.c
@@ -119,9 +119,9 @@ int write_index_parse(char *arg) {
     int fmt = HTS_FMT_CSI;
 
     if (arg) {
-        if (strcmp(arg, "csi") == 0)
+        if (strcmp(arg, "csi") == 0 || strcmp(arg, "=csi") == 0)
             fmt = HTS_FMT_CSI;
-        else if (strcmp(arg, "tbi") == 0)
+        else if (strcmp(arg, "tbi") == 0 || strcmp(arg, "=tbi") == 0)
             fmt = HTS_FMT_TBI;
         else
             return 0;
```

jkbonfield:
Ok I added -W in there too.

I'm unsure if we want to support `-W=tbi` as well as `-Wtbi`, but personally I dislike the latter and prefer the former as it shows consistency with long options and is distinct from the mandatory option versions.  Either way, it's a trivial couple lines to tweak in version.c so I'll let you decide (and a whole raft of search and replace in the docs).

Also fixed a couple other doc issues in the process.

Note I didn't change the usage statements to be more verbose.  IMO it doesn't need it, and I dislike huge long lines (some of them are 160 characters, which breaks most unix conventions).  I don't think you'd like me line-wrapping them either, as we've had that discussion in the past.  So again, if you want that doing, I'll let you make that editorial decision at merge.

mbhall88:
> It appears the frequent strategy though, if we want both long and short opts, is normally not to make them have optional args and have two options. Eg -W, --write-index plus --index-fmt FMT.

My two cents would be to prefer this approach. But I appreciate my two cents should have very low market value. Just providing a user perspective...

pd3:
> It appears the frequent strategy though, if we want both long and short opts, is normally not to make them have optional args and have two options. Eg `-W, --write-index` plus `--index-fmt FMT`.

Optional arguments are unexplored territory in bcftools and I don't know how likely they are to cause problems in practice. Adding the new `--index-fmt` option seems like a more conservative and safer approach, but I don't insist.

Issue #1322, opened 2020-10-06
bcftools compile error under fedora and centos


kelepiradam:
Hi 
I am facing an issue with fedora and centos systems when compiling bcftools with perl filters and libgsl enabled. 

`
gcc -rdynamic   -o bcftools main.o vcfindex.o tabix.o vcfstats.o vcfisec.o vcfmerge.o vcfquery.o vcffilter.o filter.o vcfsom.o vcfnorm.o vcfgtcheck.o vcfview.o vcfannotate.o vcfroh.o vcfconcat.o vcfcall.o mcall.o vcmp.o gvcf.o reheader.o convert.o vcfconvert.o tsv2vcf.o vcfcnv.o HMM.o consensus.o ploidy.o bin.o hclust.o version.o regidx.o smpl_ilist.o csq.o vcfbuf.o mpileup.o bam2bcf.o bam2bcf_indel.o bam_sample.o vcfsort.o cols.o extsort.o ccall.o em.o prob1.o kmin.o  polysomy.o peakfit.o vcfplugin.o htslib-1.11/libhts.a -lpthread -lz -lm -lbz2 -llzma -lcurl -lcrypto -lm -lz  -ldl  -lgsl -lgslcblas -Wl,--enable-new-dtags -Wl,-z,relro -Wl,--as-needed -Wl,-z,now -specs=/usr/lib/rpm/redhat/redhat-hardened-ld  -Wl,-z,relro -Wl,--as-needed -Wl,-z,now -specs=/usr/lib/rpm/redhat/redhat-hardened-ld  -fstack-protector-strong -L/usr/local/lib  -L/usr/lib64/perl5/CORE -lperl -lpthread -lresolv -ldl -lm -lcrypt -lutil -lc -lpthread
gcc   -o test/test-regidx test/test-regidx.o regidx.o htslib-1.11/libhts.a -lpthread -lz -lm -lbz2 -llzma -lcurl -lcrypto -lpthread -lz  -ldl 
/usr/bin/ld: main.o: relocation R_X86_64_32 against `.rodata.str1.1' can not be used when making a PIE object; recompile with -fPIE
/usr/bin/ld: vcfindex.o: relocation R_X86_64_32 against `.rodata.str1.8' can not be used when making a PIE object; recompile with -fPIE
/usr/bin/ld: tabix.o: relocation R_X86_64_32 against `.rodata.str1.1' can not be used when making a PIE object; recompile with -fPIE
[...]

PR #2138, opened 2024-03-25
Remove -specs=...redhat-hardened... from Perl options

jmarshall:
As encountered in #1322, Red Hat's `ExtUtils::Embed` returns CLAGS & LDFLAGS that produce a PIE position-independent executable. However libhts.a and the rest of bcftools's *.o files are not compiled as PIC, so linking fails as these other object files use relocations that are invalid for PIE.

While bcftools could link against libhts.so and compile the rest of its objects with `-fpic`, the logistics are non-trivial. And frankly it's not really reasonable for `ExtUtils::Embed` to insist that other applications follow Perl/Red Hat's policy re PIE. So it's easier to omit the redhat-hardened-cc1 and redhat-hardened-ld specs that set up for building a PIE executable, and there is already precedent in _configure.ac_ to filter undesired options out of the `ExtUtils::Embed` output.


pd3:
Thank you

PR #2166, opened 2024-04-18
Genotype Probability (GP) option for the dosage plugin

PaulTimmers:
The dosage plugin is useful for quickly converting VCF genotypes into dosages; however, to date there was no option to convert genotype probabilities into dosages. This pull request contains the addition of a single option (GP) to the dosage plugin:

bcftools +dosage input.vcf -- -t GP


pd3:
This was closed without explanation. I am assuming that's because the `+tag2tag` plugin was discovered which has this functionality.

PR #2219, opened 2024-07-04
Fix bug in vcfnorm's strdup_alt_svlen()

jmarshall:
Read SVLEN as four bytes of `int32_t` rather than just one byte. Previously the returned string's `%d` was formatted with a value representing only the first 8 bits of the full 32-bit SVLEN value.

Fixes a bug recently introduced in 6d53540567e0241ebb53c723e1352c71d571ca46. This was returning strings like `<DEL>.24` instead of `<DEL>.-1000` on little endian machines. On big-endian machines, it was returning `<DEL>.255` for all of -1000, -2000, -3000, leading to test case failures.


pd3:
Thank you, the array is defined as `uint8_t *tmp_arr1` indeed

PR #2255, opened 2024-08-13
Improve documentation of mpileup adjust-MQ.

### LINKED ISSUE #2217: How does mpileup adjust-MQ actually work?
Hi there, first of all, thanks for making bcftools! It's really amazing to have these kind of tools available.

I find the documentation around the mpileup option --adjust-MQ a bit lacking though and the option seems to sometimes give unexpected results.  
 1. Just from the documentation, I would expect the adjusted mapping quality to follow the given formula:
$adjust = \sqrt{\frac{option - quality}{option}}\cdot option$, 
which would result in the following relationship between mapping quality and adjusted mapping quality (assuming that mapping quality values larger than the value given are left as is):
adjusted_mapping quality
this however seems counterintuitive, as it would boost really bad aligning reads and reduce the value for better aligning ones (or is this what is intended?)

 2. Looking at the code it seems like bcftools calculates the mapping quality used for the adjustment independently of the reported mapping quality from the aligner (see https://github.com/samtools/htslib/blob/develop/realn.c#L39 which is what is called in mpileup). I have not looked at this specifically in illumina data, but I have an example in 10x data where all reads at a position have a reported mapping quality of 60, but some of them seem to get adjusted to below 20 (using -C 50 -q 20), judging from the resulting DP from mpileup. It seems like there is no documentation on how this quality is calculated (and I have not managed to fully understand it from the code) and for which type of data this is applicable.

I'm happy to contribute to the documentation to make this option a bit clearer once I actually understand how it works.

jkbonfield:
Note the formatting isn't great.  I don't understand how to get docbook to do tables, and even verbatim examples are generated incorrect man pages.  I'll let the finessing of formatting be done by someone who understands docbook encoding.

However the text seems OK to me.  It's the same as the samtools PR.

Fixes #2217

### DIFF
--- a/doc/bcftools.txt
+++ b/doc/bcftools.txt
@@ -2190,10 +2190,56 @@ multiple regions and many alignment files are processed.
     by misalignments.
 
 *-C, --adjust-MQ* 'INT'::
-    Coefficient  for  downgrading mapping quality for reads containing
-    excessive mismatches. Given a read with a phred-scaled probability q of
-    being generated from the mapped position, the new mapping quality is
-    about sqrt((INT-q)/INT)*INT. A zero value (the default) disables this functionality.
+    Coefficient for downgrading mapping quality for reads containing
+    excessive mismatches.  Mismatches are counted as a proportion of the
+    number of aligned bases ("M", "X" or "=" CIGAR operations), along with
+    their quality, to derive an upper-bound of the mapping quality.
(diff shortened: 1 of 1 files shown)


pd3:
Thank you, formatting fixed and merged. (By the way, the format is asciidoc and we use asciidoctor to do the formatting.)

PR #2261, opened 2024-08-19
Ensure clean compilation with -Wformat -Wformat=2

daviesrob:
Companion to samtools/htslib#1816, which among other things adds printf format checking attributes to `bcf_hdr_printf()`.  This fixes a few places that could trigger warnings in bcftools as a result of that change.  It also adds checking for some functions internal to bcftools so `clang -Wformat -Wformat=2` works and fixes a couple of bugs that were found as a result.

Bugs fixed:

 * In `open_file()` a user-controlled string was passed as the format to `mkdir_p()`, which could lead to information leakage or possibly arbitrary writes via a crafted file path name.

 * Use of wrong type for position in vcfsort `sort_blocks()` error report.

 * Printing the wrong variable in vcfsort `do_partial_merge()` error report.

The `-Wformat=2` option disallows non-literals in printf-like functions, which can be a little difficult to work around.  Here most cases were fairly easy to sort out, apart from `hdr_append()` in `plugins/fill-tags.c`.  For that the chosen solution was to convert the function into a macro. While not particularly nice it does allow all of the format strings passed in to be validated properly.

Adds `Wformat` to clang CI build to catch any future regressions.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -84,6 +84,7 @@ ubuntu_task:
 
   environment:
     CC: clang
+    CFLAGS: -g -O2 -Werror -Wall -Wformat -Wformat=2
     LC_ALL: C
     CIRRUS_CLONE_DEPTH: 1
     HTSDIR: ./htslib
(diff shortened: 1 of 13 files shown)


pd3:
Thank you

PR #2265, opened 2024-08-21
Fix a regression with -Ou format being the same as -Ob.

### LINKED ISSUE #2263: Incorrect compression on `view -Ou -o foo.bcf`
https://github.com/samtools/bcftools/commit/30dbc3c39ed92d3617fb0737ba533b1c97c3907e added code to override the requested format based on the filename suffix.  Unfortunately `-Ob` and `-Ou` now both operate the same as it claims bcf is `FT_BCF | FT_GZ`.

Hence this regressed between 1.11 and 1.12.

We can still keep the autodetection, but we also need to explicitly override clevel for -u so it's subsequently defined as level 0. Eg:

```
diff --git a/vcfview.c b/vcfview.c
index 58063ebb..63ce5438 100644
--- a/vcfview.c
+++ b/vcfview.c
@@ -629,9 +629,9 @@ int main_vcfview(int argc, char *argv[])
             case 'O':
                 switch (optarg[0]) {
                     case 'b': args->output_type = FT_BCF_GZ; break;
-                    case 'u': args->output_type = FT_BCF; break;
+                    case 'u': args->output_type = FT_BCF; args->clevel = 0;break;
                     case 'z': args->output_type = FT_VCF_GZ; break;
-                    case 'v': args->output_type = FT_VCF; break;
+                    case 'v': args->output_type = FT_VCF; args->clevel = 0;break;
                     default:
                     {
                         args->clevel = strtol(optarg,&tmp,10);
```

Note this is still subtly different.  1.11 onwards treated `-Ou` as a naked BCF with no gzip wrapper.  `-Ob -l0` however is BGZF wrapped with no-compression, but it does mean it has CRC checking.  I think the former was not spec compliant anyway as BGZF is mandatory, so this regression could be considered to be a bug fix.

jkbonfield:
Commit 30dbc3c broke the -Ou option when a filename is given. So `bcftools sort -Ou -o out.bcf in.vcf` is treated as `-Ob` instead.

Hence this was a regression in functionality that arrived with bcftools 1.12.  The fix is simply to honour Ou vs Ob even when filenames are specified, and only use the filename suffix when it's incompatible with the data type specified.

Fixes #2263

### DIFF
--- a/version.c
+++ b/version.c
@@ -88,7 +88,8 @@ void set_wmode(char dst[8], int file_type, const char *fname, int clevel)
     const char *end = fname ? strstr(fname, HTS_IDX_DELIM) : NULL;
     if ( !end ) end = fname ? fname + strlen(fname) : fname;
     int len = end - fname;
-    if ( len >= 4 && !strncasecmp(".bcf",fname+len-4,4) ) ret = hts_bcf_wmode(FT_BCF|FT_GZ);
+    if ( len >= 4 && !strncasecmp(".bcf",fname+len-4,4) )
+        ret = hts_bcf_wmode(file_type & FT_BCF ? file_type : FT_BCF|FT_GZ);
     else if ( len >= 4 && !strncasecmp(".vcf",fname+len-4,4) ) ret = hts_bcf_wmode(FT_VCF);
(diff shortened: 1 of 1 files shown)


pd3:
Thank you

PR #2272, opened 2024-09-02
Miscellaneous minor header and dependency fixes

jmarshall:
Fix _Makefile_ dependencies, and add `#include <strings.h>` where POSIXly-correctly needed for `strcasecmp()` et al.


pd3:
Thank you

PR #2281, opened 2024-09-17
Remove the mpileup BAM_CREF_SKIP filter.

### LINKED ISSUE #2277: Significant difference between IDV and AD when calling certain RNA-seq indels with mpileup
I'm using mpileup to compute the mutation frequencies of a list of known variant sites in RNA-seq. For a small number of indels the AD numbers in the output are much lower than what I'm seeing in IGV. For instance, when I run the following command:
`bcftools mpileup -r 11:69587265 -f <GRCm38 fasta> --annotate FORMAT/AD,FORMAT/DP,INFO/AD -F 0.001 --max-depth 10000 --max-idepth 10000 -Q20 -x -A --no-BAQ --tandem-qual 10000 <RNA-seq bam> | grep -v "^#"`

Here is the output I'm getting:
`11      69587265        .       CA      CAA     0       .       INDEL;IDV=77;IMF=0.616;DP=124;AD=47,2;I16=24,23,1,1,1880,75200,80,3200,872,17236,40,800,1057,25169,46,1066;QS=0.908467,0.0915332;VDB=0.56;SGB=-0.453602;RPBZ=-0.571191;MQBZ=2.56432;MQSBZ=0;BQBZ=0;SCBZ=-1.87886;MQ0F=0  PL:DP:AD        0,106,55:49:47,2`

And here is how this site looks like in IGV:
296fe156-fbce-4e9d-b37c-38660d014890

I am aware that some low quality reads are filtered during variant calling and the reported AD number will be lower than the raw IDV. However upon manual examination most of the reads seem to be high-quality and correctly aligned. Such a huge drop from 77 to 2 seems counter-intuitive to me and I wonder if this is the expected behavior of mpileup. 

(One thing I notice is that most of the read pairs containing the insertion are spliced except 2, which happens to be the same as the AD reported)

jkbonfield:
Mpileup removes alignments using the cigar ref skip operator ("N"). This was originally added in 2011 in samtools/samtools#d1643d6 with the commit message of "fixed a bug in indel calling related to unmapped and refskip reads".

Unfortunately I don't know what that bug was, but removing the code shows it still works (at least for some data!).  We need better understanding of what's going on and why it was added, so perhaps we should add a command line option to control this instead?

Fixes #2277

### DIFF
--- a/bam2bcf_edlib.c
+++ b/bam2bcf_edlib.c
@@ -1559,18 +1559,9 @@ int bcf_edlib_gap_prep(int n, int *n_plp, bam_pileup1_t **plp, int pos,
                     }
                 }
 
-                int qbeg, qpos, qend, tbeg, tend, kk;
+                int qbeg, qpos, qend, tbeg, tend;
                 uint8_t *seq = bam_get_seq(p->b);
-                uint32_t *cigar = bam_get_cigar(p->b);
                 if (p->b->core.flag & BAM_FUNMAP) continue;
-
-                // FIXME: the following loop should be better moved outside;
-                // nonetheless, realignment should be much slower anyway.
-                for (kk = 0; kk < p->b->core.n_cigar; ++kk)
-                    if ((cigar[kk]&BAM_CIGAR_MASK) == BAM_CREF_SKIP)
-                        break;
-                if (kk < p->b->core.n_cigar)
-                    continue;
 
                 // determine the start and end of sequences for alignment
                 int left2 = left, right2 = right;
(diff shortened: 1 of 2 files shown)


pd3:
I believe the edited BAM_CREF_SKIP checking code https://github.com/samtools/bcftools/blob/ef8b974dee6d3bc9f56e39a75997a64a13cbcaa1/bam2bcf_indel.c#L855-L859 simply skips these reads from realignment. If we should merge any version of this, we'd need to include a small test to demonstrate where it can be helpful.

PR #2283, opened 2024-09-18
Make --write-index work for concat with vcf (needs htslib#1839)

### LINKED ISSUE #2246: write-index does not appear to work from concat command
I was concatting 29 uncompressed vcfs and indexing the output
```
bcftools concat -o autosomes.vcf.gz --write-index {1..29}.vcf
```

However, the resultant index is more or less empty, and running `bcftools index -n autosomes.vcf.gz` returns 0.
Deleting that index and re-indexing (either via `tabix -p vcf` or `bcftools index`) produces a larger index that returns the correct number of variants from the previous command.

This seems to also occur if the input to concat is compressed, or with `--write-index=tbi`, so seems to be general issue rather than a specific combination of events.

This is with 
```
bcftools 1.20-45-g5e7ea8a0-dirty
Using htslib 1.20-54-gfbe5ff6c-dirty
```

Best,
Alex

whitwham:
This change makes concat --write-index work with uncompressed vcf as an input.

Fixes #2246.  Needs /samtools/htslib#1839 to work.

### DIFF
--- a/vcfconcat.c
+++ b/vcfconcat.c
@@ -105,9 +105,12 @@ static void init_data(args_t *args)
             error("Different number of samples in %s. Perhaps \"bcftools merge\" is what you are looking for?\n", args->fnames[i]);
 
         int j;
-        for (j=0; j<bcf_hdr_nsamples(hdr); j++)
+        for (j=0; j<bcf_hdr_nsamples(hdr); j++) {
+            fprintf(stderr, "%s %s\n", args->out_hdr->samples[j], hdr->samples[j]);
+
             if ( strcmp(args->out_hdr->samples[j],hdr->samples[j]) )
                 error("Different sample names in %s. Perhaps \"bcftools merge\" is what you are looking for?\n", args->fnames[i]);
+        }
 
         if ( args->phased_concat )
         {
@@ -257,7 +260,6 @@ static void destroy_data(args_t *args)
     free(args->fnames);
 }
 
-int vcf_write_line(htsFile *fp, kstring_t *line);
 
 #define SWAP(type_t, a, b) { type_t t = a; a = b; b = t; }
 static void phase_update(args_t *args, bcf_hdr_t *hdr, bcf1_t *rec)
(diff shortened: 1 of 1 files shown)


vcfconcat.c
```diff
@@ -683,7 +686,23 @@ static void concat(args_t *args)
...
+                        str = end + 1; // should be the beginning of the ref
+                        while (*str && *str != '\t') str++; // Get past the ID entry
+                        char *beg = str + 1;
+                        str++;
+
+                        while (*str && *str != '\t') {
+                            str++; // Get past the REF entry
+                        }
+
+                        if (vcf_write_line_with_index(args->out_fh, args->out_hdr, &fp->line, chr_id, tmp.s, pos, (str - beg)) < 0) {
```
jkbonfield, line 702: I'm not sure if this matters or not for something deliberately optimised like concat, but there is some oddity in VCF here.

The internal data structures have:

```
typedef struct bcf1_t {
    hts_pos_t pos;  // POS
    hts_pos_t rlen; // length of REF
```

On the face of it, that leads us all to assume rlen is simply reference length and can be gleaned trivially from strlen(ref) after parsing.  However I don't think that's always true and I think the header file documentation is outdated.

Specifically if there is an INFO/END tag it can override rlen, as seen here:  https://github.com/samtools/htslib/blob/develop/vcf.c#L3601-L3614

The `vcf_parse_info` function is quite heavyweight and totally goes against the grain of a fast concat loop doing nothing more than read/writes, but then again if we want to get an index which is the same that we'd get from `bcftools index` then I think that's what we have to do.  Sadly!  I suppose minimally we could do a strstr to look for `END=` and if it's found anywhere then do full parsing to check if it's END and not something else and to parse it.  It's a lot more work though.

Maybe @pd3 can advise.

I know this data exists in theory, but is it the sort of thing that only occurs in very specific examples (eg structural variation callers or gVCF), and if so would it be acceptable to simply document that `--write-index` is only valid for data from SNP and short-indel callers?

whitwham:
Okay, keeping it working close to the original is not really a viable option.  I will make a new PR with a different method.

PR #2289, opened 2024-09-26
CI improvements

jkbonfield:
- Improve cirrus-CI for shared htslib/bcftools PRs.   If the first line of the PR commit contains htslib#NUM then the htslib PR of that number is checked out when building bcftools.  This allows for testing of the same branch in both (which happens automatically for a user's fork, but not once PRs are made)

- Replace AppVeyor with GitHub actions for speed and consistency with samtools.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -7,6 +7,7 @@ version: 'vers.{build}'
 branches:
     except:
       - gh-pages
+      - /.*/ # Appveyor builds are currently disabled.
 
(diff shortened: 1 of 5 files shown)


daviesrob:
Rebased with a couple of minor adjustments:
* Restored Appveyor config, but disabled on all branches.  This makes for a cleaner transition.
* Keep `text` attribute on `.c`, `.h` and `.pl` files under `test` so they behave the same as source files elsewhere.

PR #2367, opened 2025-02-18
Add support for inverting phase of genotypes with setGT plugin

rwk-unil:
Hello,
I added an option to invert the phase of diploid genotypes in the setGT plugin.
So `0|1` becomes `1|0`.

This is useful when one wants to invert the whole phase of a file, or selected samples, or based on a filter.

The function only inverts the phase of diploid genotypes (checks `ngt==2` and if `bcf_int32_vector_end` is present).
Phase flag is set correctly per BCF specs. Note unphased genotypes are also swapped so `0/1` becomes `1/0`.

I hope we can get this merged, let me know if there are any changes you would like me to make.

Best regards,
Rick


plugins/setGT.c
```diff
@@ -113,6 +114,7 @@ const char *usage(void)
         "           X       .. allele with bigger read depth as determined from FMT/AD\n"
         "           p       .. phase genotype (0/1 becomes 0|1)\n"
         "           u       .. unphase genotype and sort by allele (1|0 becomes 0/1)\n"
+        "           i       .. inverse phase of genotype (0|1 becomes 1|0)\n"
```
pd3, line 117: I am not a native speaker, but I think a better description would be "invert the genotype phase"

plugins/setGT.c
```diff
@@ -364,6 +367,21 @@ static inline int unphase_gt(int32_t *ptr, int ngts)
     return changed;
 }
 
+// inverse phase for a single sample, ngts is the ploidy
```
pd3, line 370: invert, not inverse

pd3:
Please add a test, see `test/test.pl`, and fix the language, then it can be merged.

pd3 (approved):

pd3:
Thank you

PR #2374, opened 2025-02-28
Fix broken tests due to MSYS2 changes.

whitwham:
Due to changes in how MSYS2 perl reported the identity of the OS it was built for, our tests were failing to adapt to the Windows style file locations.

Background:
https://www.msys2.org/news/#2025-02-14-moving-msys2-closer-to-cygwin

MSYS2 PR:
msys2/MSYS2-packages#5177


pd3:
Thank you

PR #2503, opened 2026-02-12
Fix silent truncation when synced reader encounters I/O errors

gileshall:
## Summary

`bcf_sr_next_line()` returns 0 on both EOF and error, and 14 of 16 bcftools commands using the synced reader never check `files->errnum` after the read loop. When an I/O error occurs (e.g. an HTTP connection drop or cloud storage throttle), these commands interpret the 0 return as normal EOF, produce silently truncated output, and exit with code 0. There is no error message and no indication that anything went wrong.

This is the root cause of spurious truncations observed during periods of high GCS load. The companion htslib PR (samtools/htslib#1987) adds retry/resilience so most transient errors never reach callers, but when retries are exhausted the error must not be swallowed.

Add post-loop `errnum` checks to all 14 affected commands so that unrecoverable read errors produce an error message and non-zero exit code.

## Affected commands
`stats`, `query`, `filter`, `plugin`, `cnv`, `isec`, `gtcheck`, `merge`, `concat`, `convert`, `annotate`, `norm`, `csq`, `consensus`

## Already safe (no change needed)
`view`, `call`

### DIFF
--- a/consensus.c
+++ b/consensus.c
@@ -411,6 +411,7 @@ static bcf1_t **next_vcf_line(args_t *args)
         }
         return &args->files->readers[0].buffer[0];
     }
+    if ( args->files->errnum ) error("Error: %s\n", bcf_sr_strerror(args->files->errnum));
     return NULL;
 }
 static void unread_vcf_line(args_t *args, bcf1_t **rec_ptr)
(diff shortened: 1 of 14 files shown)


pd3:
Thank you

PR #2519, opened 2026-03-26
Fix off-by-one in HMM forward-backward posterior indexing

sirus20x6:
## Summary
- Fix `fwd[i*2 + state]` to `fwd[(i+1)*2 + state]` in `vcfroh.c` and equivalent in `vcfcnv.c` — the `fwd[0..nstates-1]` slot contains initial state probabilities, not site 0 posteriors
- Fix `hmm_set_tprob()` in `HMM.c` to reallocate `tprob_arr` when called with a larger `ntprob`, preventing heap buffer overflow

## Test plan
- [x] Existing test suite passes (updated roh.1.{2,3,4}.out to match corrected posteriors)
- [ ] Verify ROH quality scores are more accurate with corrected posteriors

### DIFF
--- a/HMM.c
+++ b/HMM.c
@@ -215,11 +215,12 @@ void hmm_reset(hmm_t *hmm, void *_snapshot)
 
 void hmm_set_tprob(hmm_t *hmm, double *tprob, int ntprob)
 {
+    int old_ntprob_arr = hmm->ntprob_arr;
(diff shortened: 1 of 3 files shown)


pd3:
Thank you for this. Accepted with some changes via https://github.com/samtools/bcftools/commit/60fad50b24ae1fa4a65a18079fb21ac1eee500d0

PR #2520, opened 2026-03-26
Fix vcfmerge format field handling for Number=G, INT8 PL, and dst buffer

sirus20x6:
## Summary
- Fix `le_to_i16` to `le_to_i32` in `BCF_BT_INT32` case of `merge_localized_numberG_format_field` — was reading only 2 of 4 bytes
- Add bounds check for INT8 PL values before indexing into `pl2prob[]`, matching existing INT16/INT32 checks
- Fix `src[j] = bcf_int32_vector_end` to `dst[j]` in `update_local_alleles` — was writing to source buffer instead of destination

## Test plan
- [x] Existing test suite passes (updated merge.LPL.3.out for correct missing values)
- [ ] Verify merged Number=G FORMAT fields with 32-bit integer data

### DIFF
--- a/vcfmerge.c
+++ b/vcfmerge.c
@@ -1653,7 +1653,7 @@ void init_local_alleles(args_t *args, bcf1_t *out, int ifmt_PL)
         }
         switch (fmt_ori->type)
         {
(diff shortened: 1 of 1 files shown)


pd3:
Thank you

PR #2524, opened 2026-03-26
Fix vcfannotate type mismatch, null deref, and stale buffer

sirus20x6:
## Summary
- Fix `bcf_get_info_float` to `bcf_get_info_int32` in `setter_ARinfo_int32` — was reading integer data through the float API, corrupting values
- Add NULL check for `tab` in `setter_info_int` and `setter_info_real` merge-flush paths — uses `line->n_allele`/`line->d.allele` when `tab` is NULL for Number=A/R tags
- Fix `determine_ploidy` in `vcf_setter_format_real` to use float buffers instead of stale int buffers

## Test plan
- [x] Existing test suite passes (updated annotate8.out and annotate15.out)
- [ ] Verify annotation of Number=A/R integer INFO fields and float FORMAT fields with haploid samples

### DIFF
--- a/vcfannotate.c
+++ b/vcfannotate.c
@@ -785,7 +785,7 @@ static int setter_ARinfo_int32(args_t *args, bcf1_t *line, annot_col_t *col, int
     if ( !map ) error("REF alleles not compatible at %s:%"PRId64"\n", bcf_seqname(args->hdr,line),(int64_t) line->pos+1);
 
     // fill in any missing values in the target VCF (or all, if not present)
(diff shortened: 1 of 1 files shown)


pd3:
This is not a good fix, it introduces a bug, casting forcibly `float*` to `int*` when calling `determine_ploidy()`

pd3:
This has been differently by 449a8af03f276f652519fea6b5a626b1b13000db and https://github.com/samtools/bcftools/commit/da315c5f0c8f6924d9e0b9ffc67ce2d17dce10b1

PR #2526, opened 2026-03-26
Fix literal string matching for external annotation filter values

### LINKED ISSUE #2506: Bcftools annotate does not treat INFO field containing pipes (|) as literal string when used as additional matching key
Hi,
I am encountering unexpected behaviour when using `bcftools annotate` with an additional `INFO` field as a matching key.

## Background

I decomposed and normalised a multi-allelic VCF using `bcftools norm` with the `–old-rec-tag` option, specifying the tag name `SOURCE_RECORD`. After normalisation, some variants become identical in terms of `CHROM`, `POS`, `REF` and `ALT`. However, they remain distinguishable by `SOURCE_RECORD` because they originate from different sources.

I then generated a TSV file of per-variant metrics from the normalised VCF. I want to annotate the original VCF with these metrics using `bcftools annotate`. Because there are duplicate `CHROM`, `POS`, `REF`, `ALT` entries after normalisation, I followed the approach suggested in [issue #2151](https://github.com/samtools/bcftools/issues/2151), where an additional `INFO` field can be used as a matching key to disambiguate records.

## Expected behaviour

When including `SOURCE_RECORD` as an additional key, I expect `bcftools annotate` to:

- Match records using `CHROM`, `POS`, `REF`, `ALT` and `SOURCE_RECORD`
- Treat `SOURCE_RECORD` as a literal string
- Only annotate records where the full `SOURCE_RECORD` value matches exactly

## Observed behaviour

`SOURCE_RECORD` values have the format `CHROM|POS|REF|ALT|USED_ALT_INDEX`. When this field this is used as an additional key, `bcftools annotate` does not appear to treat it as a strict literal string. Instead, it behaves as though the pipe characters are interpreted as OR separators. As a result:

- Records are matched if any substring between pipes matches
- The first duplicate variant receives the correct annotations
- Subsequent duplicates (same `CHROM`, `POS`, `REF`, `ALT` but different `SOURCE_RECORD`) incorrectly receive the annotations from the first occurrence, leading to incorrect assignment of metrics.

**Example VCF:**
```
#CHROM  POS     ID      REF     ALT     QUAL    FILTER  INFO
chr5    123456       .       C       CT      35.2    .       SOURCE_RECORD=chr5|123456|C|A,CT,G|2
chr5    123456       .       C       CT      21.6    .       SOURCE_RECORD=chr5|123457|TA|AA,CA,TGA|3
```
**Example TSV:**
```
CHR     POS     REF     ALT     VARIANT SOURCE_RECORD   METRIC1        METRIC2
chr5    123456       C       CT      chr5:123456-C/CT     chr5|123456|C|A,CT,G|2      1.0     700
chr5    123456      C       CT       chr5:123456-C/CT     chr5|123457|TA|AA,CA,TGA|3  0.0     70
```
**Annotation command:**
```
bcftools annotate ${vcf} \
    --annotations ${tsv} \
    --columns CHROM,POS,REF,ALT,-,SOURCE_RECORD,METRIC1,METRIC2 `
    --include 'SOURCE_RECORD={SOURCE_RECORD}' \
    --keep-sites \
    --header-lines ${header}
```
**Resulting VCF:**
```
#CHROM  POS     ID      REF     ALT     QUAL    FILTER  INFO
chr5    123456       .       C       CT      35.2    .       SOURCE_RECORD=chr5|123456|C|A,CT,G|2;METRIC1=1.0;METRIC2=700
chr5    123456       .       C       CT      21.6    .       SOURCE_RECORD=chr5|123457|TA|AA,CA,TGA|3;METRIC1=1.0,METRIC2=700
```
Note the second variant is assigned the wrong metrics.

If I remove the pipe characters in `SOURCE_RECORD`, the annotation behaves as expected and matching is correct, suggesting the issue is specifically related to how `bcftools annotate` interprets pipe characters in `INFO` fields used as matching keys.

# Questions

- Is there a way to force `bcftools annotate` to treat the `INFO` field as a strict literal string when used as a key?
- Alternatively (or additionally), would it be possible to make the format of `–old-rec-tag`customisable (for example, allowing the delimiter to be specified by the user)?

Thank you for your time.

Lisa

sirus20x6:
## Summary
- When `bcftools annotate` uses a filter expression with external annotation values (`{VARIABLE}` syntax), the string comparison incorrectly split values on commas and matched any component — values like `chr5|123456|C|A,CT,G|2` would match on substring `CT`
- Use direct whole-string comparison when either operand is an external value

Fixes #2506

## Test plan
- [ ] Verify annotation matching with INFO values containing commas and pipes

### DIFF
--- a/filter.c
+++ b/filter.c
@@ -2883,12 +2883,24 @@ static void cmp_vector_strings(token_t *atok, token_t *btok, token_t *rtok)
             return;
         }
         if ( !regex )
-            rtok->pass_site = _match_vector_strings(atok->str_value.s, atok->str_value.l, btok->str_value.s, btok->str_value.l, logic, missing_logic);
+        {
+            // When either operand is an external value (set via filter_test_ext),
+            // compare as literal strings rather than splitting on commas. External
+            // values come from annotation file columns and may contain commas that
+            // are part of the value, not VCF multi-value separators. (see gh #2506)
+            if ( atok->iext || btok->iext )
+            {
+                int match = atok->str_value.l==btok->str_value.l && !strncmp(atok->str_value.s,btok->str_value.s,atok->str_value.l) ? 1 : 0;
+                if ( logic==TOK_NE ) match = match ? 0 : 1;
+                rtok->pass_site = match;
+            }
+            else
+                rtok->pass_site = _match_vector_strings(atok->str_value.s, atok->str_value.l, btok->str_value.s, btok->str_value.l, logic, missing_logic);
+        }
         else
         {
             token_t *tok = atok->regex ? btok : atok;
             rtok->pass_site = _regex_vector_strings(regex, tok->str_value.s, tok->str_value.l, logic, missing_logic);
-    fprintf(stderr,"pass=%d [%s]\n",rtok->pass_site,tok->str_value.s);
         }
         return;
     }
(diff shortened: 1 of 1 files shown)


pd3:
This needs a test case to demonstrate the problem. The referenced issue #2506 could not be reproduced thus far.

Issue #2393, opened 2025-04-15
bcftools norm convert non-ATGCN IUPAC to N however mpileup and call does not.


gokalpcelik:
Hi

I noticed that mpileup and call incorporates non-ATGCN IUPAC bases into VCF however bcftools norm expects only ATGCN. Is this an expected outcome? I thought non-ATGCN bases are not allowed in VCF spec? Latest spec dictates conversion of R to A for example though 4.2 does not mention it. 

This happens when GIAB reference or any of the GATK resource bundle references are used for mapping and calling.

pd3:
What version of bcftools are you using? I believe recent versions of `bcftools mpileup` convert non-ACGT characters to N

pd3:
Please do provide a test. I tested myself and it is working as expected https://github.com/samtools/bcftools/commit/0bce7ab4832b7d3228637f25417f2aeb4646a944

sirus20x6:
Related fix in #2527 — the csq codon translation macros now clamp the index for non-ACGT bases (returning 'X' instead of reading past the gencode buffer).

PR #2527, opened 2026-03-26
Fix csq consequence prediction: sref offset, OOB codon, stop codon, splice

### LINKED ISSUE #2478: bcftools csq throw sanity_check_ref error and could not pass by --force
Hi,

I kept on getting this error when running bcftools csq and it could not go past it even when supplied with --force. Could you help?

bcftools: csq.c:2801: sanity_check_ref: Assertion `vcf - rec->d.allele[0] < strlen(rec->d.allele[0]) && ref - TSCRIPT_AUX(tr)->ref < tr->end - tr->beg + 2*N_REF_PAD' failed.
Aborted

Thanks a lot!

sirus20x6:
## Summary
- Fix trailing N_REF_PAD copy offset in spliced reference — was copying from beginning of last CDS exon instead of after it
- Clamp codon index to 0-63 in dna2aa/cdna2aa macros — non-ACGT bases (N) produced index 84, reading past the 64-char gencode string
- Fix stop codon search comparing against `tseq.l` instead of `tseq_stop.l` (2 occurrences)
- Fix `test_splice` checking hardcoded `allele[1]` instead of loop variable `allele[i]`
- Fix missing pos assignment for reverse-strand HAP_SSS consequences

Fixes #2478. Relates to #2393.

## Test plan
- [ ] Verify csq with transcripts containing N bases in reference
- [ ] Verify csq --force no longer crashes on sanity_check_ref assertion

### DIFF
--- a/csq.c
+++ b/csq.c
@@ -570,10 +570,13 @@ const uint8_t cnt4[] =
     4,3,4,2, 4,4,4,1, 4,4,4,4, 4,4,4,4,
     4,4,4,4, 0
 };
-#define dna2aa(x)    gencode->code[  nt4[(uint8_t)(x)[0]]<<4 |  nt4[(uint8_t)(x)[1]]<<2 |  nt4[(uint8_t)(x)[2]] ]
-#define cdna2aa(x)   gencode->code[ cnt4[(uint8_t)(x)[2]]<<4 | cnt4[(uint8_t)(x)[1]]<<2 | cnt4[(uint8_t)(x)[0]] ]
-#define dna2stop(x)  gencode->stop[  nt4[(uint8_t)(x)[0]]<<4 |  nt4[(uint8_t)(x)[1]]<<2 |  nt4[(uint8_t)(x)[2]] ]
-#define cdna2stop(x) gencode->stop[ cnt4[(uint8_t)(x)[2]]<<4 | cnt4[(uint8_t)(x)[1]]<<2 | cnt4[(uint8_t)(x)[0]] ]
+#define _codon_idx(a,b,c) ((a)<<4 | (b)<<2 | (c))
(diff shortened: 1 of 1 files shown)


csq.c
```diff
@@ -2208,7 +2211,7 @@ void hap_add_csq(args_t *args, hap_t *hap, hap_node_t *node, int tlen, int ibeg,
         }
         for (i=0; i<hap->tseq_stop.l; i++)
             if ( hap->tseq_stop.s[i]=='*' ) break;
-        if ( i!=hap->tseq.l )
+        if ( i!=hap->tseq_stop.l )
```
pd3, line 2214: This change is not needed, tseq_stop and tseq have always the same length

csq.c
```diff
@@ -2292,6 +2295,7 @@ void hap_add_csq(args_t *args, hap_t *hap, hap_node_t *node, int tlen, int ibeg,
 
     if ( hap->stack[ibeg].node->type == HAP_SSS  )
     {
+        node->csq_list[icsq].pos          = hap->stack[ibeg].node->rec->pos;
```
pd3, line 2298: Also this change is not needed, pos is always set when the list is extended.

csq.c
```diff
@@ -2955,7 +2959,7 @@ int test_cds_local(args_t *args, bcf1_t *rec)
                 }
                 for (j=0; j<tseq_stop->l; j++)
                     if ( tseq_stop->s[j]=='*' ) break;
-                if ( j!=tseq->l )
+                if ( j!=tseq_stop->l )
```
pd3, line 2962: As above

pd3:
Thank you for this. Not all of the suggested changes were meaningful, so I committed them separately.

PR #2528, opened 2026-03-26
Fix vcfconvert ALT parsing, null deref, and sample compression

sirus20x6:
## Summary
- Fix ALT allele scanning in `_set_chrom_pos_ref_alt` — while loop checked `*tsv->se` (constant end pointer) instead of `*se` (scan pointer), making it a no-op
- Add NULL guard before `bgzf_thread_pool(hout, ...)` when hap output is suppressed with `.`
- Fix `sample_compressed = 0` to `= 1` in three places — .gz sample filenames were never actually compressed

## Test plan
- [ ] Verify CHROM:POS_REF_ALT format parsing in convert
- [ ] Verify `.gz` sample output files are compressed


pd3:
Thank you

PR #2529, opened 2026-03-26
Fix vcfcnv OOB read with control sample and AA skip mismatch

sirus20x6:
## Summary
- Guard `parse_lrr_baf` for control sample when `idx=-1` (no control configured) — was reading BAF/LRR at negative array offset
- Initialize `baf2`/`lrr2` to safe defaults when control sample parse is skipped
- Add `baf > 4/5.` skip (AA genotypes) in the third optimization loop to match the first two loops — without it, loop counter `k` exceeded `tmpf` bounds

## Test plan
- [ ] Verify `bcftools cnv` without control sample does not crash

### DIFF
--- a/vcfcnv.c
+++ b/vcfcnv.c
@@ -871,8 +871,9 @@ static int update_sample_args(args_t *args, sample_t *smpl, int ismpl)
     for (i=0; i<args->nsites; i++)
     {
         float baf = smpl->baf[i];
+        if ( baf>4/5.) continue;        // skip AA genotypes
         if ( baf>0.5 ) baf = 1 - baf;   // the bands should be symmetric
-        if ( baf<1/5.) continue;        // skip RR,AA genotypes
+        if ( baf<1/5.) continue;        // skip RR genotypes
 
         double prob_cn3 = args->tmpf[k++];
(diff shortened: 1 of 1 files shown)


vcfcnv.c
```diff
@@ -871,8 +871,9 @@ static int update_sample_args(args_t *args, sample_t *smpl, int ismpl)
     for (i=0; i<args->nsites; i++)
     {
         float baf = smpl->baf[i];
+        if ( baf>4/5.) continue;        // skip AA genotypes
         if ( baf>0.5 ) baf = 1 - baf;   // the bands should be symmetric
-        if ( baf<1/5.) continue;        // skip RR,AA genotypes
+        if ( baf<1/5.) continue;        // skip RR genotypes
```
pd3, line 876: This is not required. In fact, it does not change the behavior. But that's mathematically speaking - maybe the folding could in principle classify some edge cases differently, then k's could differ in the first and the third loop... OK

pd3:
Fixed via https://github.com/samtools/bcftools/commit/99f0afb10868c16cf93794ddbd92002897651251

PR #2530, opened 2026-03-26
Fix ccall type-punning and mcall division by zero

sirus20x6:
## Summary
- Fix `tmpi` declaration from `float` to `int32_t` in `ccall.c` — was passed to `bcf_update_info_int32` which interpreted float bit pattern as garbage integer for CLR tag
- Add zero-denominator guard in `mcall.c` before dividing by sum of DP4 counts — produces `mq = 0` when depth is zero, matching the pattern in `test16_core`

## Test plan
- [ ] Verify CLR INFO tag contains correct integer values after consensus calling


pd3:
Thank you

PR #2531, opened 2026-03-26
Fix bam2bcf indel calling: memset destroying consensus, uninit variable, wrong sizes

sirus20x6:
## Summary
- Fix `memset(aux, 0, ...)` to `memset(cns, 0, ...)` in `bam2bcf_iaux.c` — was zeroing nucleotide frequency counts just computed, causing the consensus builder to discard all insertion data
- Fix `memset(read_scores, 0, n)` to `memset(read_scores, 0, n * sizeof(int))` — only zeroed 1/4 of the int array
- Fix per-sample indel fraction denominator from `naux` (cross-sample running total) to `ntot` (per-sample count)
- Initialize uninitialized variable `K` in `bam2bcf_indel.c`
- Remove duplicate `memset(call->SCR, ...)` in `bam2bcf.c`

## Test plan
- [ ] Verify indel calling with `--indels-2.0` produces correct insertion consensus sequences


pd3:
Good catch, thank you

PR #2532, opened 2026-03-26
Fix tabix NULL deref on open failure and dead detect=0 code

sirus20x6:
## Summary
- Add NULL check on `bgzf_open` return in the read-all branch — was guaranteed to crash if the file could not be opened
- Move `detect = 0` from inside the error `else` block (dead code after `return 1`) to after the type-matching block — `-p` preset was being overridden by filename auto-detection

## Test plan
- [ ] Verify `bcftools tabix -p bed foo.vcf.gz` uses BED format, not VCF


pd3:
Thank you

PR #2533, opened 2026-03-26
Fix regidx out-of-bounds array access and double free

sirus20x6:
## Summary
- Cap `iend` to `list->nidx - 1` instead of `list->nidx` in `regidx_overlap` — the loop `for (i=ibeg; i<=iend; i++)` could read `list->idx[nidx]`, one past the allocated array
- Set `str.s = NULL` after `free(str.s)` in `regidx_init` — prevents double free if `hts_close()` fails and the error path frees `str.s` again

## Test plan
- [ ] Verify region overlap queries at chromosome boundaries

### DIFF
--- a/regidx.c
+++ b/regidx.c
@@ -289,6 +289,7 @@ regidx_t *regidx_init(const char *fname, regidx_parse_f parser, regidx_free_f fr
     }
 
     free(str.s);
+    str.s = NULL;
     if ( hts_close(fp)!=0 )
     {
         fprintf(stderr,"[%s] Error: close failed .. %s\n", __func__,fname);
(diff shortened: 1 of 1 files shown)


pd3:
Thank you

regidx.c
```diff
@@ -441,7 +442,7 @@ int regidx_overlap(regidx_t *regidx, const char *chr, uint32_t beg, uint32_t end
         if ( !i )
         {
             int iend = iBIN(end);
-            if ( iend > list->nidx ) iend = list->nidx;
+            if ( iend >= list->nidx ) iend = list->nidx - 1;
```
pd3, line 445: @daviesrob If correct, I think this affects also regidx.c code in htslib

daviesrob: I'm not convinced that this is correct.  The item count [gets incremented by 1](https://github.com/samtools/bcftools/blob/0039c202c99afd222367c0898aaf2c28b7f76d6b/regidx.c#L396) when allocating `list->idx`, so it can accommodate looking up `iend`.  I assume that it was written that way because the actual range is likely to end somewhere in the middle of the bin.

Because the array is big enough, there is no out of bounds access.

PR #2534, opened 2026-03-26
Fix vcfstats per-sample statistics using wrong FORMAT column index

### LINKED ISSUE #2469: bcftools stats --samples-file give wrong results if sample order is different than vcf
I'm using bcftools v1.21

If I run bcftools stats with --sample-files, I've noticed that the per-sample statistics will be incorrect if the sample order in the input list of samples does not match the order of the samples in the input vcf. The statistics in the output file will keep the original order in the vcf file, but the order of the names will match the order in the input file.

sirus20x6:
## Summary
- When `--samples-file` is used and sample order differs from the VCF, FORMAT field accessors (`calc_sample_depth`, `get_ad`, `get_iad`) used the subset index `is` instead of the VCF column index `reader->samples[is]`
- Introduced `ismpl = reader->samples[is]` and used it consistently for all FORMAT data access

Fixes #2469

## Test plan
- [ ] Verify `bcftools stats -s sample_list` with reordered sample list

### DIFF
--- a/vcfstats.c
+++ b/vcfstats.c
@@ -1103,8 +1103,10 @@ static void do_sample_stats(args_t *args, stats_t *stats, bcf_sr_t *reader, int 
     int is;
     for (is=0; is<args->files->n_smpl; is++)
     {
+        int ismpl = reader->samples[is];    // VCF column index for this sample
+
         // Determine depth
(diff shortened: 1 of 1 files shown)


pd3:
This is now added, with a test. Thank you

PR #2535, opened 2026-03-26
Replace assert() with proper error handling for user-triggerable conditions

sirus20x6:
## Summary
Several `assert()` calls guard conditions that can be triggered by user input. In release builds with NDEBUG, these are compiled out, causing silent undefined behavior or buffer overflows.

- **consensus.c**: Replace 3 asserts (variant overlap, fasta buffer) with `error()` calls
- **convert.c**: Replace 2 asserts (GT type check) with if-checks that warn and skip; add 2 allele index bounds checks before `line->d.allele[ial]` access
- **vcfconcat.c**: Replace assert on bgzf block size with `error()` call
- **filter.c**: Add NULL check on `getenv("HOME")` before passing to `kputs`

## Test plan
- [ ] Verify consensus/convert handle malformed input gracefully instead of crashing

### DIFF
--- a/consensus.c
+++ b/consensus.c
@@ -815,7 +815,9 @@ static void apply_variant(args_t *args, bcf1_t *rec)
         // forward
         int ntrim = args->fa_frz_pos - rec->pos + 1;
         int nref  = strlen(rec->d.allele[0]);
-        assert( ntrim < nref );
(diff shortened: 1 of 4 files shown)


pd3:
Thank you

PR #2536, opened 2026-03-26
Fix missing header finalization and unchecked faidx return

sirus20x6:
## Summary
- Add missing `bcf_hdr_add_sample(args->header, NULL)` in `gensample_to_vcf` (`vcfconvert.c`) — header was not finalized after adding samples, matching the pattern in `haplegendsample_to_vcf`
- Add NULL check on `faidx_fetch_seq` return in `indel_ctx_type` (`vcfstats.c`) — prevents NULL dereference when chromosome is not in the reference

## Test plan
- [ ] Verify `bcftools convert --gensample2vcf` with sample files
- [ ] Verify `bcftools stats --ref-file` with mismatched reference

### DIFF
--- a/vcfconvert.c
+++ b/vcfconvert.c
@@ -489,6 +489,7 @@ static void gensample_to_vcf(args_t *args)
         *se = 0;
         bcf_hdr_add_sample(args->header,samples[i]);
     }
+    bcf_hdr_add_sample(args->header,NULL);
     for (i=0; i<nsamples; i++) free(samples[i]);
     free(samples);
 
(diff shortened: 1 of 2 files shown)


vcfconvert.c
```diff
@@ -489,6 +489,7 @@ static void gensample_to_vcf(args_t *args)
         *se = 0;
         bcf_hdr_add_sample(args->header,samples[i]);
     }
+    bcf_hdr_add_sample(args->header,NULL);
```
pd3, line 492: This is no longer required, see comments in vcf.c

pd3:
Resolved via 8cd70c911d622dcd4efcee3364e455d1a4b50b07, the change in vcfconvert.c is not required.

Issue #2557, opened 2026-05-16
`view -R`: pathological performance with large single-base region sets (84M regions, 11+ hours unfinished)


carstenerickson:
## Summary

`bcftools view -R sites.bed.gz target.vcf.gz` is 350-400× slower than a sequential scan when the BED contains many densely-clustered single-base regions — a common pattern in PRS / ancestry pipelines (HGDP+1kGP, AADR 1240k, PGS Catalog). A production run on 84M such regions against a 23 GB VCF did not complete in 11+ hours of 100% CPU before we killed it.

## Reproducer

```bash
python3 gen_reproducer.py --regions 1000000 --prefix synth1m
bgzip -f synth1m.bed && tabix -p bed synth1m.bed.gz
bgzip -f synth1m.vcf && tabix -p vcf synth1m.vcf.gz
bcftools view -R synth1m.bed.gz synth1m.vcf.gz > /dev/null
```

<details>
<summary><code>gen_reproducer.py</code> (N densely-clustered single-base positions, written to both BED and VCF)</summary>

```python
import argparse, random

# Approx GRCh38 autosome lengths (Mbp) — used as relative weights only.
CHROM_LEN_MBP = [248,242,198,190,181,170,159,145,138,133,
                 135,133,114,107,101, 90, 83, 80, 58, 64, 46, 50]

def gen_positions(n_total, avg_spacing, seed=42):
    rng = random.Random(seed)
    rate = 1.0 / avg_spacing
    total = sum(CHROM_LEN_MBP)
    for chrom_i, mbp in enumerate(CHROM_LEN_MBP, start=1):
        length, share = mbp * 1_000_000, n_total * mbp // total
        pos = 0
        for _ in range(share):
            pos += max(1, int(rng.expovariate(rate)))
            if pos >= length: break
            yield str(chrom_i), pos

ap = argparse.ArgumentParser()
ap.add_argument("--regions", type=int, default=1_000_000)
ap.add_argument("--avg-spacing", type=int, default=10)
ap.add_argument("--prefix", default="synth")
args = ap.parse_args()
positions = list(gen_positions(args.regions, args.avg_spacing))

with open(f"{args.prefix}.bed", "w") as f:
    for chrom, pos in positions:
        f.write(f"{chrom}\t{pos-1}\t{pos}\n")

with open(f"{args.prefix}.vcf", "w") as f:
    f.write("##fileformat=VCFv4.2\n")
[...]

PR #2561, opened 2026-05-16
view: opt into htslib dense single-base BED auto-promotion (#2557)

carstenerickson:
## Summary

Wires `bcftools view -R FILE` into the new `BCF_SR_AUTO_TARGETS_FROM_REGIONS` htslib opt (samtools/htslib#2011). When `-R FILE` is given and no `-T` is set, view sets the opt before `bcf_sr_set_regions()`; htslib's sniffer decides per-file whether to promote a dense single-base BED to the streaming-targets path.

## Design

| condition                        | behaviour |
|----------------------------------|-----------|
| `-R FILE`, no `-t/-T`            | opt on (auto) |
| `-R FILE -T OTHER`               | opt off (avoid API state leak: sniffer would populate `readers->targets`, conflicting with `set_targets`) |
| `-r REGION` (string)             | opt off (sniffer requires a file) |
| `-R FILE --no-regions-fastpath`  | opt off (user escape hatch) |

`--no-regions-fastpath` is the user-facing escape hatch for corner cases the sniffer accepts incorrectly.

## Measurements

End-to-end via `bcftools view -R`; single-base BED at 10 bp avg spacing matching the VCF positions; macOS arm64:

|     N | upstream 1.23.1 | this PR | speedup |
|------:|----------------:|--------:|--------:|
|  100K |       11.7 s    | 0.17 s  |    69× |
|    1M |      156.6 s    | 0.73 s  |   215× |

The 1M case puts production HGDP+1kGP / AADR / PGS panel intersections into the minutes-not-hours regime via the most common entry point.

## Test plan

`test_vcf_view_regions_fastpath` (`test/test.pl`) asserts byte-identical output across three paths over a 300-entry single-base BED fixture:

- `view -R FILE` (fastpath; opt on)
- `view -R FILE --no-regions-fastpath` (slow path preserved)
- `view -T FILE` (control)

Existing tests pass unchanged.

## Notes

Depends on samtools/htslib#2011; without that PR, the opt has no effect and view falls back to the slow path with no semantic change.

Tracks #2557.


carstenerickson:
CI fixes pushed (force-push, post-rebase):

- **Rebased onto current `develop`** to resolve the NEWS conflict; my `bcftools view` entry now sits under `## Release a.b` in the alphabetic position after `+trio-dnm3`. No other diff vs. the previous push — the auto-merged `vcfview.c` and `test/test.pl` chunks were textually clean and the regression test still passes locally.

- **The Cirrus `'BCF_SR_AUTO_TARGETS_FROM_REGIONS' undeclared` errors are expected** until [samtools/htslib#2011](https://github.com/samtools/htslib/pull/2011) merges. This PR's `vcfview.c` references the new opt added there; the bcftools submodule pin on `develop` points at upstream htslib which doesn't yet have the enum. The natural sequence is: land htslib#2011 → bump bcftools' htslib submodule pin → merge this PR. Happy to refresh the submodule bump as part of this PR once htslib#2011 is in, or leave it to whoever does the routine submodule sync.

carstenerickson:
Closing in tandem with samtools/htslib#2011. This was just the bcftools side wiring for the opt added there, and Rob has indicated the multi region iterator approach in htslib supersedes both. Reproducer remains at samtools/bcftools#2557.

PR #2569, opened 2026-06-15
Fix vcfstats.c on big-endian platforms

jmarshall:
Byteswap as appropriate. As mentioned in PR #2044, this may avoid problems with unaligned access on little-endian platforms too.

After application of this PR, bcftools's tests pass on s390x.

Note that there still exist numerous `#define BRANCH…` macros that do not have a `convert` parameter. These presumably represent functionality that is not big-endian-correct but which is not exercised by the test suite.


pd3:
Thank you

PR #2570, opened 2026-06-16
Ensure PACKAGE_VERSION is set in the Makefile

daviesrob:
In git repositories, this is set by expanding `$(shell git describe --always --dirty)`.
As this is not in a normal Makefile rule, failure just results in an empty variable.  To check that it really was set, add a check to ensure it's not empty in the rule that builds the `version.h` file.

See also https://github.com/samtools/samtools/issues/2337 where failing to set this variable resulted in broken files.  It's less of a problem for bcftools due as it always [appends the string "+htslib" to its headers](https://github.com/samtools/bcftools/blob/15a40910a1624e8bd03bcd4eb7e03a2115ceddc6/vcfmerge.c#L3366), however it's probably a good idea to catch the failue to set a value if it occurs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -130,7 +130,13 @@ endif
 # version.h: force
 #	echo '#define BCFTOOLS_VERSION "`git describe --always --dirty`"' > $@
 version.h:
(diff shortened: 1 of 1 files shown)


pd3:
Thank you

PR #2575, opened 2026-07-03
Make reglist_sort_() stable

daviesrob:
Add a tie-breaker comparison so that `reglist_sort_()` preserves the original ordering when regions compare exactly.  This only matters where the region list includes payload data for the regions being sorted, otherwise region entries that tie are indistinguishable.  As in that case an array of pointers into the original region list is sorted, the pointer values can be used to break any ties.

This resolves a test failure for the `bcftools csq --greedy` option on alpine Linux because its qsort() implementation did not give the same ordering as the glibc one.  As glibc often (but not always!) [uses mergesort](https://elixir.bootlin.com/glibc/glibc-2.43.9000/source/stdlib/qsort.c#L394) for its qsort() implementation, it may have given the impression of being stable when in fact this is not guaranteed.

### DIFF
--- a/regidx.c
+++ b/regidx.c
@@ -143,7 +143,14 @@ static int cmp_reg_ptrs(const void *a, const void *b)
 }
 static int cmp_reg_ptrs2(const void *a, const void *b)
 {
-    return cmp_regs(*((reg_t**)a),*((reg_t**)b));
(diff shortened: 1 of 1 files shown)


pd3:
Thank you

PR #2579, opened 2026-07-07
Make gtcheck use a stable sort for "idbl"

jkbonfield:
Qsort isn't always a stable sort.  Use the array index for ties. (We could use the pointers themselves, but the struct already contains an index.)

Fixes a test issue on OpenSolaris, and probably Alpine (although it looks to have gotten lucky.)


pd3:
Thank you

PR #2590, opened 2026-09-10
csq: fix segfault when the VCF header has no contig lines

balwierz:
`bcftools csq` segfaults on a VCF whose header has no `##contig` lines — in init, before any record is read, so even a header-only VCF crashes it.

```
printf '##fileformat=VCFv4.2\n#CHROM\tPOS\tID\tREF\tALT\tQUAL\tFILTER\tINFO\n' > empty.vcf
bcftools csq -f ref.fa -g annot.gff3.gz empty.vcf   # Segmentation fault
```

In `init_chr_names()`, `bcf_hdr_seqnames()` returns `nseq=0` and a non-NULL `calloc(0,…)` pointer (glibc), so the `!vcf` guard misses and `vcf[0]` / `strcmp()` read past a zero-length array. Fix also guards `nseq==0`.


pd3:
Thank you

PR #2592, opened 2026-09-14
Switch containers to ones built by samtools/c-maint

daviesrob:
Containers ghcr.io/samtools/*-for-htslib-ci:latest include the dependencies needed to build HTSlib, SAMtools and BCFtools. They are also tested to ensure that they can build the latest release.

This removes the need to install packages on each CI build, reducing the load on upstream repositories, and making it less likely that builds will fail for reasons unrelated to the code changes that are being tested.


pd3:
Thank you

